PR #33, opened 2021-09-07
tuto getting started: modif typo + better frame slicing example

LucBourrat1:
Some typo modifications
The example taken for the frame slicing was not working well because the slice was made outside of the image size. It is modified


thibo73800:
Please provide on description on the changes you made. This is an important detail for keeping track of all the changes for the next release note.

Issue #32, opened 2021-09-07
wheel outside of the bounding box when slicing on an example frame


LucBourrat1:
for the first image of the sample cocodataset, when cropping the corresponding frame and viewing the cropped frame, a wheel gets outside of the bounding box.

Code too reproduce the bugg:
coco_dataset = CocoDetectionDataset(sample=True)
frame = coco_dataset.getitem(0)
frame.get_view().render()
cropped_frame = frame[:,100:500,200:400]
cropped_frame.get_view().render()

PR #34, opened 2021-09-07
Prevent crop outside of the spatial size

thibo73800:
In this merge request:
- Prevent the user from cropping outside of the spatial size of the current augmented tensor.


thibo73800:
Closes https://github.com/Visual-Behavior/aloception/issues/32

PR #36, opened 2021-09-07
3 cocodetection mask implement

Johansmm:
New features:  
* Include `Segementation` and `append_segmentation` attribute/function in `Frame` class
* `Segementation` attribute in `Frame`: To work with segmentation datasets by binary features maps.
   * Update `Mask` class Include `get_view` function to render segmentation 
* Use `return_mask` propierty of `COCODetectionDataset` to return a `segmentation` tensor.
* Implement `stuff_ann_file` attribut of `COCODetectionDataset` to provide the stuff annotation json file (complement the things classes).
* Integrate stuff classes and thing classes in `COCODetectionDataset`


alodataset/coco_detection_dataset.py
```diff
@@ -16,7 +17,7 @@
 from torchvision.datasets.coco import CocoDetection
 
 from alodataset import BaseDataset
-from aloscene import BoundingBoxes2D, Frame, Labels
+from aloscene import BoundingBoxes2D, Frame, Labels, Panoptic
```
thibo73800, line 20: Wonder if `Panoptic` is the right class name. Panoptic refer to the task it self (stuff + things) while this tensor actually refer to the data.  In imgaug, they refer to this type of class as : Segmentation Maps. Maybe `Segmentation` instead of `Panoptic` ? What do you think ?

alodataset/coco_detection_dataset.py
```diff
@@ -88,13 +92,18 @@ def __init__(
             dataset_dir = BaseDataset.get_dataset_dir(self)
             img_folder = os.path.join(dataset_dir, img_folder)
             ann_file = os.path.join(dataset_dir, ann_file)
+            stuff_ann_file = None if stuff_ann_file is None else os.path.join(dataset_dir, stuff_ann_file)
```
thibo73800, line 95: Does the original detr paper use the panoptic ann file or the stuff file ? Using the stuff ann combined with the things ann file is an option. But we might considere using the panoptic file as well to match the original DETR implementation.

alodataset/coco_detection_dataset.py
```diff
@@ -171,10 +184,12 @@ def getitem(self, idx):
             names=("N", None),
             labels=labels_2d,
         )
-
-        # frame.append_boxes2d(boxes.xcyc().rel_pos())
         frame.append_boxes2d(boxes)
 
+        if self.prepare.return_masks:
+            mask = Panoptic(target["masks"], names=("C", "H", "W"), labels=labels_2d)
```
thibo73800, line 190: If my understanding is correct the structure is the following:
Frame
    - Boxes 2D
        - label
    - Panoptic
        - label

There is therefore a constraints panoptic.shape[0] must be equal to boxes.shape[0] right ? 
Is there any reason why you choose to not use one mask per boxes as you previously mentioned ?

Let's talk about it during the day

alodataset/coco_detection_dataset.py
```diff
@@ -282,15 +298,14 @@ def show_random_frame(coco_loader):
 def main():
     """Main"""
     logging.basicConfig(
-        level=logging.INFO,
-        format="[%(asctime)s][%(levelname)s] %(message)s",
-        datefmt="%d-%m-%y %H:%M:%S",
+        level=logging.INFO, format="[%(asctime)s][%(levelname)s] %(message)s", datefmt="%d-%m-%y %H:%M:%S",
```
thibo73800, line 301: Its getting weird to use logger in this main() method while we don't use it in other main methods in the datasets folder. I would suggest to remove it from here, removing the print, and use the sample as for the others datasets.

aloscene/panoptic.py
```diff
@@ -0,0 +1,100 @@
...
+from aloscene.renderer import View
+from aloscene.labels import Labels
+
+
+class Panoptic(aloscene.tensors.AugmentedTensor):
+    """Binary mask for each label to represent panoptic segmentations
+    """
+
+    @staticmethod
+    def __new__(cls, x: Tensor, labels: Union[dict, Labels] = None, names=("N", "H", "W"), *args, **kwargs):
```
thibo73800, line 18: We must probably describe what type of information `labels` is supposed to  store. Before to merge,I would also suggest to add the docstrings in this class.

thibo73800 (changes requested):

alodataset/coco_detection_dataset.py
Johansmm: Ok, you are right. I will change the name

alodataset/coco_detection_dataset.py
Johansmm: I will create a new issue to make this

alodataset/coco_detection_dataset.py
Johansmm: Yes. len(labels) == len(Boxes 2D) and there is a constraints with len(labels) == len(Panoptic). So  len(Panoptic) == len(Boxes 2D).

The reason is used Segementation in applications that not required boxes.

alodataset/coco_detection_dataset.py
Johansmm: Done

aloscene/panoptic.py
Johansmm: Panoptic was replaced by update of `mask` class

Johansmm:
@thibo73800 All changes were made in the update of the mask class. Ready for revision.

aloscene/mask.py
```diff
@@ -13,20 +22,113 @@ class Mask(aloscene.tensors.SpatialAugmentedTensor):
...
-    def __new__(cls, x, *args, **kwargs):
+    def __new__(cls, x, labels: Union[dict, Labels] = None, *args, **kwargs):
         # Load frame from path
         if isinstance(x, str):
             x = load_mask(x)
-            kwargs["names"] = ("C", "H", "W")
+            kwargs["names"] = ("N", "H", "W")
         tensor = super().__new__(cls, x, *args, **kwargs)
+        tensor.add_label("labels", labels, align_dim=["N"], mergeable=True)
+        tensor.add_property("mask_size", x.shape)
```
thibo73800, line 32: This line does not seem great to me. If the frame get resized, the mask will be resized and this property might not be updated properly. Also it is still possible to get the shape of the mask at any time using self.shape

aloscene/mask.py
```diff
@@ -13,20 +22,113 @@ class Mask(aloscene.tensors.SpatialAugmentedTensor):
...
+        tensor.add_property("is_mask", not isinstance(labels, Labels))
```
thibo73800, line 33: Does `is_mask` make sense for a class with the name `Mask` ? 

The following is weird
```
mask = Mask()
mask.is_mask # False
```

thibo73800 (changes requested):

aloscene/frame.py
```diff
@@ -142,6 +138,19 @@ def append_disparity(self, disparity, name=None):
         """
         self._append_label("disparity", disparity, name)
 
+    def append_segmentation(self, segmentation: Mask, name: str = None):
+        """Attach a mask to the frame.
```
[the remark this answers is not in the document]
Johansmm: Ok. I updated the info as follows:

```
 Parameters
        ----------
        segmentation: aloscene.Mask
            Mask with size (N,H,W), where N is the features maps, each one for one object.
            Each feature map must be a binary mask. For that, is a type of aloscene.Mask
```

aloscene/mask.py
Johansmm: Done. Property removed.

aloscene/mask.py
Johansmm: OK. I removed this property also

Johansmm:
@thibo73800 Ready for review again.

thibo73800 (approved):
Sounds good

PR #68, opened 2021-09-16
Run from run_ID

LucBourrat1:
when launching a train after having loaded a checkpoint, parameters of the initial model were not kept and default parameters replaced them. It's fixed.
new parse argument "--nostrict" used if we want to run a model from a checkpoint, with different coefficients between the model and the model that generated the checkpoint.

### DIFF
--- a/alonet/common/pl_helpers.py
+++ b/alonet/common/pl_helpers.py
@@ -71,6 +71,7 @@ def add_argparse_args(parent_parser, add_pl_args=True, mode="training"):
     parser.add_argument("--project_run_id", type=str, help="Project related with the run ID to load")
     parser.add_argument("--expe_name", type=str, default=None, help="expe_name to be logged in wandb")
     parser.add_argument("--no_suffix", action="store_true", help="do not add date suffix to expe_name")
+    parser.add_argument("--strict", action="store_false", help="load from checkpoint to run a model with different weights (default True)")
 
     return parent_parser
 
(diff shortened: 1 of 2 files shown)


alonet/common/pl_helpers.py
```diff
@@ -71,6 +71,7 @@ def add_argparse_args(parent_parser, add_pl_args=True, mode="training"):
     parser.add_argument("--project_run_id", type=str, help="Project related with the run ID to load")
     parser.add_argument("--expe_name", type=str, default=None, help="expe_name to be logged in wandb")
     parser.add_argument("--no_suffix", action="store_true", help="do not add date suffix to expe_name")
+    parser.add_argument("--nostrict", action="store_false", help="load from checkpoint to run a model with different weights (default True)")
```
thibo73800, line 74: Usually if one use --nostrict the expected behavior is : Do not strictly load the weights (store true on nostrict)
But here --nostrict set nostrict to false. It seem to me to be counter intuitive.

thibo73800 (changes requested):
Beyound my last comment, it seem to me that some changes are missing for this pull request to be completed

thibo73800 (approved):

alonet/common/pl_helpers.py
LucBourrat1: now if --nostrict is specified (store True and False by default) it is possible to load a checkpoint with different keys in the load_state_dict. BUT it is still not possible for a checkpoint with same keys but different values (weights dimensions).

thibo73800 (approved):

PR #73, opened 2021-09-20
42 development panoptic module

### LINKED ISSUE #42: Development module to train panoptic head
* Development an alonet module to implement panoptic head 
* Implement criterion/losses, following original paper
* Implement inference function
* Implement train procedure, with possibility of frozen DETR weights.

Note:
* #3  is required.

Johansmm:
Development of full Panoptic Head pipeline for [DETR](https://arxiv.org/pdf/2005.12872.pdf) models:  

* Create detr_panoptic alonet module:
   * Architecture based on [original repository](https://github.com/facebookresearch/detr/blob/main/models/segmentation.py): **`PanopticHead`**, that use the modules `MHAttention` and `FPN-style`.  
      * Forward/inference optimized: only pass through the network the valid bounding boxes. In inference, the masks are aligned according to their respective boxes, giving some of them zero-masks if there is no correspondence. Finally, the inference ensures that each predicted pixel has only one category, according to the highest score detected.
   * New `PanopticCriterion`, that use `DetrPanoptic`, adding the `loss_mask` function
   * `DetrHungarianMatcher` is used as matcher Panoptic module
   * `LitPanopticDetr `class, module to train the `PanopticHead`. The module uses the matcher in train step to reduce the masks predicted in forward procedure, and default "get_outs_dilters" in valid step. Also, the default model, criterion, matcher and callbacks functions were defined. Finally, **the default model is able to use the [original pre-trained weights](https://dl.fbaipublicfiles.com/detr/detr-r50-panoptic-00ce5173.pth)**.
   * `PanopticObjectDetectorCallback`, the callback that implements the correct way to get the `pred_boxes` and `pred_masks` from inference.
   * New/draft version of Panoptic Quality (PQ) metric, with its corresponding callback. **NOT IMPLEMENTED YET IN PIPELINE**

* Also, new implicit functions were implemented/improved:
  * **New outputs in DETR/DEFORMABLE DETR forward:** backbone outputs and encoder outputs, used them in panoptic forward.
  * IoU metric between masks.
  * Convert a set of masks in id-class matrix (panoptic view).
  * Update of `CocoDetectionDataset` and `CocoDetection2Detr` modules, in order to read an "extra" json annotation file, with stuff categories. Also, the `labels_names` in `CocoDetectionDataset` were fixed to correspond to [pre-trained weights](https://dl.fbaipublicfiles.com/detr/detr-r50-panoptic-00ce5173.pth) of the original repository, through the new parameter: `fix_classes_len`. Also, remove `CATEGORIES` useless parameter. 
  * New `CocoPanoptic2Detr` dataset, as well as the  `LitPanopticDetr` data connector. This dataset use the panoptic annotation (as original repository). Also, `SplitMix` module is used to define the paths.
  * Update `DetrCriterion`, adding the `loss_mask` function, based on `DICE/F-1` and` focal` losses.
  * Update APmetrics, with the Iou between masks for `APmask` metric, and split inference function from main process in order to inherit it and modify just the inference procedure.
  * Implement mask view in `wandb.Image` method.

* Finally, some problems were fixed:  
  * `SpatialAugmentedTensor` resize operation when tensor is empty (bug)
  * `aloscence.Mask` getview for empty masks
  * If `limit_train_batches` flag is used in train procedure with and value less than 1000, no metrics were displayed by the training loop on wandb interface.
  * Add wandb to requirements.txt file
  * boxes id out of range due to removal of background class in `boxes_to_wandb` function

* Known problems:  
  *  To show the masks in wandb, the panoptic view implement a format of numpy.uint8, that only enables 256 classes. For more amount of classes, this map must be updated to RGB channel.

Closes #42


alonet/detr/coco_data_modules.py
```diff
@@ -157,6 +157,7 @@ def val_check(self):
         )
         self.sample = self.coco_val.sample or self.sample  # Update sample if user prompt is given
         self.CATEGORIES = self.coco_val.CATEGORIES if hasattr(self.coco_val, "CATEGORIES") else None
+        self.labels_names = self.coco_val.labels_names if hasattr(self.coco_val, "labels_names") else None
```
thibo73800, line 160: what is the difference between CATEGORIES & labels_names ?

alonet/detr/detr.py
```diff
@@ -31,6 +31,8 @@ def __init__(
         aux_loss=True,
         weights: str = None,
         return_dec_outputs=False,
+        return_enc_outputs=False,
+        return_bb_outputs=False,
```
thibo73800, line 35: For consistency between detr & deformable, we have to make sure theses will also be include in deformable. (pull master before to do it sine there some change that might be relevant for you)

alonet/panoptic/callbacks.py
```diff
@@ -0,0 +1,46 @@
+import aloscene
```
thibo73800, line 1: we can probably rename panoptic to detr_panoptic

alonet/panoptic/criterion.py
```diff
@@ -0,0 +1,187 @@
...
+            List of tuple with predicted indices and target indices
+        num_boxes: torch.Tensor
+            Number of total target boxes
+
+        Returns
+        -------
+        Dict
+            Losses of the loss procedure.
+        """
+        loss_map = {"labels": self.loss_labels, "boxes": self.loss_boxes, "masks": self.loss_masks}
```
thibo73800, line 185: Would be better to surcharge instead of inherit this method so that this criterion could be plug & play on different criterion following the same style.

alonet/panoptic/matcher.py
```diff
@@ -0,0 +1,133 @@
+# Mostly copy past from https://github.com/facebookresearch/detr
```
thibo73800, line 1: So this file will be removed right ?

alonet/panoptic/panoptic.py
```diff
@@ -0,0 +1,217 @@
...
+        dec_outputs, filters, gmq_params = get_filter_fn(frames=frames, m_outputs=out, **kwargs)
+
+        # FIXME h_boxes takes the last one computed, keep this in mind
+        # Use box embeddings as input of Multi Head attention
+        bbox_mask = self.bbox_attention(dec_outputs, out["enc_outputs"], mask=mask)
+        # And then, use MHA ouput as input of FPN-style CNN
+        seg_masks = self.mask_head(self.detr.input_proj(src), bbox_mask, [features[i][0] for i in range(3)[::-1]])
+
+        out["pred_masks"] = seg_masks.view(bs, bbox_mask.shape[1], seg_masks.shape[-2], seg_masks.shape[-1])
+        out["infer_info"] = {"frame_size": frames.shape[-2:], "gmq_params": gmq_params, "filters": filters}
```
thibo73800, line 125: to rename pred_masks_info

alonet/panoptic/panoptic.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Returns
+        -------
+        aloscene.BoundingBoxes2D
+            Boxes from DETR model
+        aloscene.Maks
+            Binary mask of PanopticHead
+        """
+
+        # Get information about boxes
```
thibo73800, line 146: would it be possible to surcharge the inference method of Detr instead of doing the boxes inference again ?

aloscene/mask.py
```diff
@@ -110,7 +97,56 @@ def __get_view__(self, title=None):
             )
         return View(frame, title=title)
 
+    def masks2panoptic(self):
```
thibo73800, line 100: Where is this method used ?

aloscene/tensors/spatial_augmented_tensor.py
```diff
@@ -475,6 +475,9 @@ def _resize(self, size, **kwargs):
         h = self.relative_to_absolute(size[0], "h", assert_integer=False)
         # assert_integer = False to avoid raising Assertion. Ex: when coordinates = 900.0000000000001
         w = self.relative_to_absolute(size[1], "w", assert_integer=False)
+        if ("N" in self.names and self.size("N") == 0) or ("C" in self.names and self.size("C") == 0):
```
thibo73800, line 478: Can you add some comment above the condition to explain slightly what is done in this condition ?

thibo73800 (changes requested):

alodataset/coco_detection_dataset.py
```diff
@@ -106,10 +106,9 @@ def __init__(
 
         # Setup the class names
         nb_category = max(cat["id"] for cat in cats)
+        if stuff_ann_file is not None and name == "coco":
+            nb_category = 250  # From original repo https://github.com/facebookresearch/detr/blob/main/models/detr.py
```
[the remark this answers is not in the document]
Johansmm: @thibo73800  You are right. But, without that, there is a problem in backward process. The number of classes in background does not match with the number of neurons in embed class layer. 

For this reason, I just use this number of categories for "coco" dataset and `stuff != None`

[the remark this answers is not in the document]
thibo73800: @Johansmm  You mean there is an issue if nb_category=real number of category & you use the pre-traied weights with 250 categories ? If so then nb_category should be equal to the real number of categories and then should have the possibility to be changed in the constructor to a predefined number of categories: 250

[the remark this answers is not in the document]
Johansmm: But this should go like a new parameter of COCO Datset ?

alonet/callbacks/map_metrics_callback.py
```diff
@@ -49,13 +48,26 @@ def on_validation_batch_end(
...
         if isinstance(outputs["m_outputs"], list):
             is_temporal = True
             b_pred_boxes = pl_module.inference(outputs["m_outputs"])
+            if isinstance(b_pred_boxes, tuple):
+                is_pre_mask = True
+                b_pred_masks = b_pred_boxes[1]
+                b_pred_boxes = b_pred_boxes[0]
         else:
             b_pred_boxes = [pl_module.inference(outputs["m_outputs"])]
+            if isinstance(b_pred_boxes[0], tuple):
```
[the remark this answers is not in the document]
Johansmm: Ok. I will create a "inference" function in `map_metrics_callback.py`.

alonet/callbacks/map_metrics_callback.py
```diff
@@ -49,13 +48,26 @@ def on_validation_batch_end(
...
+                is_pre_mask = True
```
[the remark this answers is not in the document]
Johansmm: Resolved in previous comment

alonet/detr/coco_data_modules.py
Johansmm: `CATEGORIES` is a set, and `label_names` is a list. For that, `CATEGORIES` keep only the unique values. But you are right. This attribute is not as important as `labels_names`. I will remove it

alonet/detr/detr.py
Johansmm: Done

alonet/panoptic/callbacks.py
Johansmm: Ok; I got it

alonet/panoptic/matcher.py
Johansmm: Yes. I will remove it

alonet/panoptic/criterion.py
Johansmm: @thibo73800  What do you mean?  You suggest update `detr_criterion`, right?

alonet/panoptic/panoptic.py
Johansmm: Done

alonet/panoptic/panoptic.py
Johansmm: @thibo73800   Before, I made this procedure. But, In term of align boxes and masks, I decided to make the boxes "again". The reasons:
* The inference need the b_filters to align the masks.
To make sure that the filters will be the same, the inference create the boxes. But if you don't agree, I can try it.

aloscene/mask.py
Johansmm: @thibo73800  In alonet.callbacks.object_detector_callback.py, on `log_masks`  function. It is a function like a `rgb2id`, but adapted to aloscene.Mask

aloscene/tensors/spatial_augmented_tensor.py
Johansmm: Done

alodataset/coco_detection_dataset.py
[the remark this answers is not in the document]
thibo73800: Yes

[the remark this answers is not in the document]
thibo73800: @Johansmm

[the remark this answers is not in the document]
Johansmm: Ok, I got it. Thanks

alonet/panoptic/panoptic.py
thibo73800: I might be wrong but I think that both the  inference of deformable and detr can use a filter in the infernece method
detr.inference(filter=my_filter). If the filter is defined, then the model will use it instead of using the default one. Let me know

Johansmm: Ok I will try it. Thanks

Johansmm:
@thibo73800 New updates in inference and loss_mask function. Now, boxes and masks are aligned. Also, some problems are fixed about backward from empty masks.

alodataset/coco_detection_dataset.py
```diff
@@ -174,6 +185,9 @@ def getitem(self, idx):
             target["boxes"] = target["boxes"][idxs]
             target["labels"] = torch.from_numpy(new_labels[idxs])
 
+            if self.prepare.return_masks:
```
thibo73800, line 188: Isn't that supposed to be in the coco_panoptic only ?

thibo73800: Let me know what you think, Just asking

alonet/common/logger.py
```diff
@@ -77,25 +77,31 @@ def log_image(trainer, key, images):
...
-        List of dict with two keys
+        List of dict with
         `image` : torch.Tensor
-        `boxes`: aloscene.BoundingBoxes2D
-    boxes: list, optional
-        List of Dict containing each two keys : name: str, boxes: aloscene.BoundingBoxes2D
+        `boxes`: aloscene.BoundingBoxes2D, optional
+        `masks`: aloscene.Mask, optional
     """
     if isinstance(trainer.logger, WandbLogger):
```
thibo73800, line 85: Did you handle how this code behave using --log tensorboard ? Its important to support it since its gonna become the default logger at some point.

alonet/detr/criterion.py
```diff
@@ -6,6 +6,67 @@
 import aloscene
 
 
+def dice_loss(inputs: torch.Tensor, targets: torch.Tensor, num_boxes: int) -> torch.Tensor:
```
thibo73800, line 9: Theses new extensions should be in the `detr_panoptic` instead of in the `detr` folder. You should have a DetrPanopticCriterion that inhrt this class

alonet/detr/train.py
```diff
@@ -207,6 +197,8 @@ def build_criterion(
         loss_ce_weight=1,
         loss_boxes_weight=5,
         loss_giou_weight=2,
+        loss_dice_weight=0,
+        loss_focal_weight=0,
```
thibo73800, line 201: Same, for me theses parameters should exists only in the detr_panoptic folder

alonet/detr_panoptic/callbacks.py
```diff
@@ -0,0 +1,46 @@
...
+        assert "m_outputs" in outputs
+
+        if isinstance(batch, list):
+            frames = aloscene.Frame.batch_list(batch)
+        else:
+            frames = batch
+
+        pred_boxes, pred_masks = pl_module.inference(outputs["m_outputs"])
+        self.log_boxes_2d(frames=frames, preds_boxes=pred_boxes, trainer=trainer, name="train/frame_obj_detector")
+        self.log_masks(frames=frames, pred_masks=pred_masks, trainer=trainer, name="train/frame_seg_detector")
```
thibo73800, line 30: TOOD: Check if it works with tensorboard (if not done already)

alonet/detr_panoptic/detr_panoptic.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(
+        self,
+        DETR_module: alonet.detr.Detr,
+        freeze_detr: bool = True,
+        aux_loss: bool = None,
+        device: torch.device = torch.device("cpu"),
+        weights: str = None,
+    ):
+        super().__init__()
+        self.detr = DETR_module
```
thibo73800, line 51: If the goal is to reuse this panoptic module for deformable as well, then self.detr could be rename by something less specific: `self.base_model` ?

alonet/detr_panoptic/detr_panoptic.py
```diff
@@ -0,0 +1,222 @@
...
+            else:
+                raise ValueError(f"Unknown weights: '{weights}'")
+
+    @assert_and_export_onnx(check_mean_std=True, input_mean_std=INPUT_MEAN_STD)
+    def forward(self, frames: aloscene.frame, get_filter_fn: Callable = None, **kwargs):
+        """PanopticHead forward, that joint to the previous boxes predictions the new masks feature.
+
+        Parameters
+        ----------
+        frames : aloscene.frame
```
thibo73800, line 87: `get_filter_fn` seem important to described

alonet/detr_panoptic/detr_panoptic.py
```diff
@@ -0,0 +1,222 @@
...
+        features = out["bb_outputs"]
+        src, mask = features[-1][0], features[-1][1].to(torch.float32)
+        mask = mask[:, 0].to(torch.bool)
+        bs = src.shape[0]
+
+        # Filter boxes and get mask indices from them
+        get_filter_fn = get_filter_fn or (lambda *args, **kwargs: get_mask_queries(*args, model=self.detr, **kwargs))
+        dec_outputs, filters = get_filter_fn(frames=frames, m_outputs=out, **kwargs)
+
+        # FIXME h_boxes takes the last one computed, keep this in mind
```
thibo73800, line 120: Is there still a fixe todo here ?

thibo73800 (changes requested):
Congreat on this work :) Minor changes to do before to merged, but we should be ready soon.
Also, you cna add your model to the main readme along with a specific readme in the detr_panoptic folder for more details.

alodataset/coco_detection_dataset.py
```diff
@@ -57,6 +56,10 @@ def __init__(
             Include masks labels in the output, by default False
         classes : list, optional
             List of classes to be filtered in the annotation reading process, by default None
+        fix_classes_len : int, optional
+            Fix to a specific number the number of classes, filling the rest with "N/A" value.
+            Use when the number of model outputs does not match with the number of classes in the dataset,
+            by default None
         stuff_ann_file: str, optional
```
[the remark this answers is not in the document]
Johansmm: @thibo73800, ummm yes. I think you are right. Maybe is very specific for COCO2017 dataset, that split thing/stuff objects annotations into two files. 

I have one idea: We could remove this parameter and maybe create a new boolean parameter called `return_multiple_labels` that enables used Labels as dictionary with all different categories (category, supercategory, crowd, isthing, ) ..., obviosly if annotation file has this properties. So, if an user want to use "thing/stuff", he must include it in annotation file. What do you thing ?
Is some similar that the thing made in CocoPanopticDataset ...

alodataset/coco_detection_dataset.py
Johansmm: @thibo73800, if annotations has "segmentation" attribute, any dataset could be represented as segmentation. For example, instance_train2017.json and instance_val2017.json coco annotations have this atribute. So, it is possible get the mask for this dataset.

alonet/common/logger.py
Johansmm: @thibo73800  Done in new commit.

alonet/detr_panoptic/callbacks.py
Johansmm: Done

alonet/detr_panoptic/detr_panoptic.py
Johansmm: @thibo73800 It is correct. But, if I change this parameter, the pre-trained weights won't match (given by the attribute name changed). So, Should I change the pre-trained weights too?

alonet/detr_panoptic/detr_panoptic.py
Johansmm: Not. Corrected

alonet/detr/criterion.py
Johansmm: Done

alonet/detr/train.py
Johansmm: All these parameters were removing

alonet/detr_panoptic/detr_panoptic.py
Johansmm: Done

alodataset/coco_detection_dataset.py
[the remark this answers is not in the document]
thibo73800: What about having something like this
    - CocoDataset : Simply ann_folder * ann_images
    - CocoDetectionDataset : will include specific is_crowd & isthing & stuff_ann_file
    - CocoPanoptic : panoptic operations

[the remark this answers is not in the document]
thibo73800: CocoDetectionDataset inhert from CocoDataset
& CocoPanoptic (Or CocoDetectionPanoptic ?) inhert from CocoDetectionDataset

alodataset/coco_detection_dataset.py
thibo73800: got it

alonet/detr_panoptic/detr_panoptic.py
thibo73800: I see, let's find a solution fo this issue once we got to adapt the code to handle both detr & Deformable

alodataset/coco_detection_dataset.py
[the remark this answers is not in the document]
Johansmm: For the moment, I think that is better remove stuff annotations and handled this new datasets in another issue.  i dont know if you agree. I think that its a new issue.

alonet/detr_panoptic/detr_panoptic.py
Johansmm: Ok. I will create a new issue to adapt panoptic head to deformable

thibo73800 (approved):

thibo73800:
@Johansmm  Does unit test passed ? If so I'll merge the branch

Johansmm:
@thibo73800  yes. All unittest pass

Issue #43, opened 2021-09-07
Development panoptic metrics


Johansmm:
Implement PQ, SQ, RQ metrics and use them in panoptic model.

PR #77, opened 2021-09-21
Coco panoptic dataset

Johansmm:
New alodataset: CocoSegmentationDataset.
Also, new data connector: CocoPanoptic2Detr


alodataset/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .split_mixin import SplitMixin
 from .waymo_dataset import WaymoDataset
 from .coco_detection_dataset import CocoDetectionDataset
+from .coco_segmentation_dataset import CocoSegementationDataset
```
thibo73800, line 6: Doest CocoPanopticDataset would be more accurate ?

alodataset/coco_segmentation_dataset.py
```diff
@@ -0,0 +1,223 @@
...
+            masks_to_boxes(masks),
+            boxes_format="xyxy",
+            absolute=True,
+            frame_size=frame.HW,
+            names=("N", None),
+            labels=labels_2d,
+        )
+        frame.append_boxes2d(boxes_2d)
+
+        if self.return_masks:
```
thibo73800, line 210: Does it make sense to not return the masks for a `Panoptic` dataset ?

alonet/panoptic/coco_data_modules.py
```diff
@@ -0,0 +1,59 @@
...
+architecture.
+"""
+
+from typing import Optional
+
+import alodataset
+import alonet
+
+
+class CocoPanoptic2Detr(alonet.detr.CocoDetection2Detr):
```
thibo73800, line 12: I think it is time to slightly organised the detr data modules into one folder:

data_modules /
    data2detr.py : include global transformations and common parameters
    coco2detr.py : include coco dataset specific methods
    cocopanoptic2detr.py include coco panoptic dataset specific methods

alodataset/coco_segmentation_dataset.py
Johansmm: @thibo73800  Maybe in the cases that we need only the boxes? and the format of json is like panoptic.json ? I dont sure

thibo73800: Ok, but let's set `return_masks` True  by default then

alodataset/__init__.py
Johansmm: OK

alonet/panoptic/coco_data_modules.py
Johansmm: Good idea. Done and tested

alodataset/coco_segmentation_dataset.py
Johansmm: Done

Johansmm:
@thibo73800 All the suggestion were made. This new dataset is a requirement of #42. Ready to merge

**Note: I include in alonet.detr.__ini__.py this modules, in order to not modify any class that depends of CocoDetection2Detr**

alodataset/coco_panoptic_dataset.py
```diff
@@ -0,0 +1,223 @@
...
+        frame.append_boxes2d(boxes_2d)
+
+        if self.return_masks:
+            masks_2d = Mask(masks, names=("N", "H", "W"), labels=labels_2d)
+            frame.append_segmentation(masks_2d)
+        return frame
+
+
+if __name__ == "__main__":
+    coco_seg = CocoPanopticDataset()
```
thibo73800, line 217: Should we upload the sample of `CocoPanopticDataset` ? and make this __main__ of this file used the sample instead of the real dataset ? (as all other datasets)

thibo73800 (changes requested):
Sounds great to me. Just a little comment to addressed before to merged about the sample

thibo73800:
One last thing I'm thinking about: We need to update the documentation to include this coco panoptic dataset in the alodataset API section

alodataset/coco_panoptic_dataset.py
Johansmm: Donde. For the moment `CocoPanopticDataset` uses the same `CocoDetectionDataset` sample. In #43 will be implement new labels for things/stuff classes. So the sample for `CocoPanopticDataset` will be update on that moment.

Johansmm:
@thibo73800  I created a new issue for that #86  Ready to merge

thibo73800 (approved):

thibo73800:
@Johansmm  Is the sample ready to be uploaded ?

Johansmm:
@thibo73800 yes. When #43 issue is solved, we have to change this sample with things/stuff new labels.

PR #80, opened 2021-09-21
update load weight and load train functions

### LINKED ISSUE #74: load weights from path/run id
Load weights in 3 different ways:
* By run id
* By weights path (.pth)
* By checkpoint path

Johansmm:
Updates:   
* `load_weight` and `load_training` functions to enable the load weights ".pth", ".ckpt" or by run id.
* load weights in detr and deformable models

Closes #74

### DIFF
--- a/alonet/common/pl_helpers.py
+++ b/alonet/common/pl_helpers.py
@@ -16,7 +16,7 @@ def vb_folder():
     alofolder = os.path.join(home, ".aloception")
     if not os.path.exists(alofolder):
         raise Exception(
-            f"{alofolder} do not exists. Please, create the folder with the appropriate files. (Checkout documentation)"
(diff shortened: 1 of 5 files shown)


thibo73800:
`load_training` can used both run_id & weights with .pth weights as well. But looking at the code it seem to me that `load_weights` in detr & deformable can't be created based on `run_id` right ?

Johansmm:
Exactly, because detr is a `nn.module` and it does not have `project_id` and `run_id. attributes`. But instead, I update `load_weights` function to load weights from .ckpt files.

thibo73800:
Ok, I guess this is not an issue right now. I'm just thinking that at some point it would be great to be able to load models directly from run_id.  https://github.com/Visual-Behavior/aloception/issues/85

thibo73800 (approved):

PR #81, opened 2021-09-21
sample download progress bar and skip user prompt

### LINKED ISSUE #75: Bar progress in samples download
Update samples download process to show its progress. Also, remove user prompt if sample does not exist in list of samples.

Johansmm:
Update `BaseDataset` to:  
* Show progress bar while a sample is downloading.
* Skip "Do you want to download and use a sample?" if `name` not in possible samples. Now, the workflow directly make the following user prompt: "Please write a new root directory for {self.name} dataset: " 

Closes #75

### DIFF
--- a/alodataset/base_dataset.py
+++ b/alodataset/base_dataset.py
@@ -4,6 +4,7 @@ import requests
 import shutil
 import torch
 import json
+from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


thibo73800:
Thanks for the feature ! However, unit test do not pass anymore using the new samples

PR #87, opened 2021-09-23
Fixe weights is None for detr & Deformable

thibo73800:
In this pull request:
- Fixe weights is None for detr & Deformable

### DIFF
--- a/alonet/deformable_detr/deformable_detr.py
+++ b/alonet/deformable_detr/deformable_detr.py
@@ -169,7 +169,7 @@ class DeformableDETR(nn.Module):
             self.to(device)
 
         if weights is not None:
-            if weights is not None and (
+            if (
                 weights in ["deformable-detr-r50", "deformable-detr-r50-refinement"]
                 or ".pth" in weights
                 or ".ckpt" in weights
--- a/alonet/deformable_detr/deformable_detr.py
+++ b/alonet/deformable_detr/deformable_detr.py
@@ -114,7 +114,8 @@ class DeformableDETR(nn.Module):
                 in_channels = backbone.num_channels[i]
                 input_proj_list.append(
                     nn.Sequential(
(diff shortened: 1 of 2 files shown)


Johansmm:
@thibo73800, Is it possible to remove the double condition "`if weights is not None"` for both cases? I mean, in the internal conditions, I think that is not necessary ask again `if weights is not None`

alonet/deformable_detr/deformable_detr.py
```diff
@@ -167,14 +168,15 @@ def __init__(
...
-        if weights is not None and (
-            weights in ["deformable-detr-r50", "deformable-detr-r50-refinement"]
-            or ".pth" in weights
-            or ".ckpt" in weights
-        ):
-            alonet.common.load_weights(self, weights, device, strict_load_weights=strict_load_weights)
-        else:
-            raise ValueError(f"Unknown weights: '{weights}'")
+        if weights is not None:
+            if weights is not None and (
```
Johansmm, line 172: Maybe is not necessary make the question "weights is not None" again

alonet/detr/detr.py
```diff
@@ -83,10 +83,11 @@ def __init__(
         if device is not None:
             self.to(device)
 
-        if weights is not None and (weights == "detr-r50" or ".pth" in weights or ".ckpt" in weights):
-            alonet.common.load_weights(self, "detr-r50", device, strict_load_weights=strict_load_weights)
-        else:
-            raise ValueError(f"Unknown weights: '{weights}'")
+        if weights is not None:
```
Johansmm, line 86: Same here

Johansmm (changes requested):

alonet/deformable_detr/deformable_detr.py
thibo73800: yes, done

alonet/detr/detr.py
thibo73800: yes, done

PR #105, opened 2021-10-11
Quick fix log mask and panoptic

Johansmm:
Three problems solved:  
* log_mask in wandb incompatible with color by object (not by ID class).
* Updata panoptic/finetune panoptic parameters, including `strict` and `base_model`(respectively)
* Update load_training and load_weights with correct extension file comparison

### DIFF
--- a/alonet/callbacks/object_detector_callback.py
+++ b/alonet/callbacks/object_detector_callback.py
@@ -27,7 +27,7 @@ class ObjectDetectorCallback(pl.Callback):
 
         Parameters
         ----------
-        val_frames: list | alonet.Frame
+        val_frames: list of :mod:`~aloscene.frame`
             List of sample from the validation set to use to load the validation progress
         one_color_per_class
             Set same segmentation-color for all objects with same category ID, by default True


alonet/callbacks/object_detector_callback.py
```diff
@@ -129,8 +130,12 @@ def log_masks(self, frames: list, pred_masks: list, trainer: pl.trainer.trainer.
             A set of predicted segmentation masks
         trainer: pl.trainer.trainer.Trainer
             Lightning trainer
+        name : str
+            Name to show in wandb
         """
         images = []
+        # Show categories with same color if there are many categories
+        color_by_cat = len(frames[0].segmentation.labels.labels_names) > 20
```
thibo73800, line 138: Sounds arbitrary. Why 20 ?

alonet/callbacks/object_detector_callback.py
```diff
@@ -142,29 +147,51 @@ def log_masks(self, frames: list, pred_masks: list, trainer: pl.trainer.trainer.
...
             frame = frame.permute([1, 2, 0]).contiguous().numpy()
 
             # Get panoptic view
-            target_masks = target_masks.mask2id()
-            p_mask = p_mask.mask2id()
-            target_masks[target_masks == -1] = len(labels_names)  # Background N/A
-            p_mask[p_mask == -1] = len(labels_names)  # Background N/A
+            target_masks = target_masks.mask2id(return_cats=color_by_cat)
+            p_mask = p_mask.mask2id(return_cats=color_by_cat)
+            if VOID_CLASS_ID < 0:
```
thibo73800, line 176: what is VOID_CLASS_ID ?

alonet/callbacks/object_detector_callback.py
Johansmm: @thibo73800  Yes. It is a value that a choose. For datasets with high amount of  objects type  (like COCO), it is better show one color by object id (all `person` with the same color) . But, if we used a finetune model for detect only persons (for example), its better show every person detected by one different color. I though that maybe this process could make automatically, selecting as threshold the value of 20

alonet/callbacks/object_detector_callback.py
Johansmm: It is a ID-constant  used by mask to set every pixel that have not prediction into a common value. This value is used in compute_pq, callbacks and aloscene.mask. For that, I preferred to save it in a common file, and used it by import

thibo73800 (changes requested):

Johansmm:
@thibo73800  Done

PR #109, opened 2021-10-15
Depth2disp2pts3d

thibo73800:
In this merge request:
- Update camera intrinsic class
- Disp2Depth
- Depth2Disp
- Points3D
- Unit tests for disp 2  depth & depth 2 pts3d


aloscene/tensors/spatial_augmented_tensor.py
```diff
@@ -12,21 +12,48 @@
...
+
+    cam_intrinsic: CameraIntrinsic
+        Camera Intrinsic. If provided, the focal_length, plane_size, principal_point, camera_side & baseline will not
+        be used. An error will be raised if theses parameters are provided since using the `cam_intrinsic` and theses
+        parameters could be ambigious.
+    cam_extrinsic: CameraExtrinsic
+        Camera extrinsic parameters as an homogenious transformation matrix.
+    camera_side : {'left', 'right', None}
+        If part of a stereo setup, will encode the side of the camere "left" or "right" or None.
+    baseline: float | None
```
ragier, line 27: Why not just only in disparity ?
It's very specific compared to SpatialAugmentedTensor which represent only grid based data

ragier: An alternative would be to keep track of the "sensor" which as produce the information, namely the frame?
But I agree it's a simpler this way.

thibo73800:
https://github.com/Visual-Behavior/aloception/pull/118

PR #128, opened 2021-11-05
Raft refacto

jsalotti:
Make code more generic and readable :

- the class of correlation block can be given as a parameter of the constructor
- flow upsampling logic is factorized in method `upsample_flow()` 
- the outputs of an update step are regrouped in a dict ; 
- at each update step, the output dict is appended to list `m_outputs` 
- the processing after last update step is factorized in `forward_head`
- `out_plane` allows to predict a map of vectors of different sizes (1D, 2D, 3D, etc.)
- added properties `hdim` and `cdim` as short_names for `hidden_dim` and `context_dim`
 
Better implementation for abstract class attributes : 
- declare abstract attribute in parent class with `my_attribute = abstract_attribute()`
- check that abstract attributes are instantiated in child class by overriding parent class `__new__` method
- two methods are provided to easily override `__new__` : `check_abstract_attribute_instanciation` and `super_new`


thibo73800:
@jsalotti can you provide a description of the changes ?

alonet/raft/raft.py
```diff
@@ -37,20 +37,23 @@ class RAFTBase(nn.Module):
     corr_levels = None
     corr_radius = None
 
+    out_plane = None
```
thibo73800, line 40: Just curious. Why corr_levels, corr-radius and out_plane are create as static None variable  ? Usually static variable are defined as constant and are != None

thibo73800, line 40: Can we add a comment to say that corr_levels, corr_radius and out_plane must be set by the child class ?

alonet/raft/raft.py
```diff
@@ -83,7 +86,7 @@ def build_update_block(self, update_cls=BasicUpdateBlock):
         """
         Build RAFT Update Block
         """
-        return update_cls(self.corr_levels, self.corr_radius, hidden_dim=self.hdim)
+        return update_cls(self.corr_levels, self.corr_radius, hidden_dim=self.hdim, out_planes=self.out_plane)
```
thibo73800, line 89: Also here, if the child class do not implement out_plane, can we make sure theses values are != None and raise an exception otherwise

thibo73800 (approved):

PR #143, opened 2021-12-14
Compatibility with embed systems

Johansmm:
Remove __future__ annotations and typing outputs to correct compatibility with Jetson Xavier


jsalotti:
Note for future development : the current python version supported on the jetson Xavier is python3.6 (see jetpack 4.6). If a more recent python version becomes supported in next releases of jetpack, then the __future__ imports should work again.

thibo73800 (approved):

PR #144, opened 2021-12-16
add_rotation

LucBourrat1:
Rotation transformation is added into transforms.py, and points2d, augmented_tensors and spatial_augmented_tensors are adapted consequently.
Spatial_shift is adapted for points2d

### DIFF
--- a/aloscene/points_2d.py
+++ b/aloscene/points_2d.py
@@ -429,6 +429,35 @@ class Points2D(aloscene.tensors.AugmentedTensor):
             abs_size = tuple(s * fs for s, fs in zip(size, points.frame_size))
             return points.abs_pos(abs_size)
 
+    def _rotate(self, angle, **kwargs):
+        """Rotate Point2d, but not their labels
+
+        Parameters
+        ----------
+        size : float
+
+        Returns
+        -------
+        points : aloscene.Point2d
+            rotated points
+        """
+        points = self.clone()
+        H, W = self.frame_size
+        angle_rad = angle * np.pi / 180
+        rot_mat = torch.tensor([[np.cos(angle_rad), np.sin(angle_rad)], [-np.sin(angle_rad), np.cos(angle_rad)]]).to(
+            torch.float32
+        )
+        tr_mat = torch.tensor([W / 2, H / 2])
+        for i in range(points.shape[0]):
+            points[i] = torch.matmul(rot_mat, points[i] - tr_mat) + tr_mat
+
+        max_size = torch.as_tensor([W, H], dtype=torch.float32)
+        points_filter = (points >= 0).as_tensor() & (points <= max_size).as_tensor()
+        points_filter = points_filter[:, 0] & points_filter[:, 1]
+        points = points[points_filter]
+
+        return points
+
     def _crop(self, H_crop: tuple, W_crop: tuple, **kwargs):
         """Crop Points with the given relative crop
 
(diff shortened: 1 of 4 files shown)


aloscene/points_2d.py
```diff
@@ -429,6 +429,35 @@ def _resize(self, size, **kwargs):
...
+        Parameters
+        ----------
+        size : float
+
+        Returns
+        -------
+        points : aloscene.Point2d
+            rotated points
+        """
+        points = self.clone()
```
thibo73800, line 444: At a first glance is seem to me that you do not check the point format (xy, or yx) neither if the points are "absolute" or "relative". Please, make sure the following is robust the differents encoding state.

aloscene/tensors/augmented_tensor.py
```diff
@@ -48,34 +48,29 @@ def __init__(self, x, **kwargs):
...
+        """Remove all children from this augmented tensor and return
         the removed children.
         """
         labels = {}
         for name in self._children_list:
-            labels[name] = {
-                "value": getattr(self, name),
-                "property": self._child_property[name]
-            }
+            labels[name] = {"value": getattr(self, name), "property": self._child_property[name]}
```
thibo73800, line 56: your ide might not be configured properly

aloscene/tensors/spatial_augmented_tensor.py
```diff
@@ -457,6 +457,24 @@ def _resize(self, size, **kwargs):
...
+        angle : float
+
+        Returns
+        -------
+        rotated : aloscene.SpatialAugmentedTensor
+            rotated tensor
+        """
+        # If a SpatialAgumentedTensor is empty, rotate operation does not work. Use view instead.
+        if ("N" in self.names and self.size("N") == 0) or ("C" in self.names and self.size("C") == 0):
+            shapes = list(self.shape)[:-2] + [h, w]
```
thibo73800, line 474: How the following works ? [h, w] is not defined right ?

thibo73800 (changes requested):

aloscene/tensors/augmented_tensor.py
LucBourrat1: I have same configuration as the one which is recommanded in the aloception gitlab.

aloscene/points_2d.py
LucBourrat1: right

thibo73800: Very late reply : looking back at this code the points format is not checked. Therefore you'll rotate based on W and H assuming the points are encoded as "xy" but this might not be the case.

thibo73800 (changes requested):

aloscene/points_2d.py
```diff
@@ -441,7 +441,7 @@ def _rotate(self, angle, **kwargs):
         points : aloscene.Point2d
             rotated points
         """
-        points = self.clone()
+        points = self.xy().clone()
```
thibo73800, line 444: clone is redundant , xy() is already cloning the vector.

thibo73800 (changes requested):

thibo73800:
Once this last comment is sloved it should be good for me to merge this branch into master

aloscene/points_2d.py
LucBourrat1: done

thibo73800 (approved):

PR #145, opened 2022-01-11
Panoptic2trt

Johansmm:
Export **`PanopticHead` in `TensorRT `,** in one/two engine(s) with dynamic axes.

**New features:**
 * Update `detr_panoptic()`  forward to enable exportation (avoid not ops. implementation on tensorRT 8.2.0.6)
 * Implement `PanopticTRTExporter` class: Base exporter for Panoptic-based modules. According its configuration, it will export full model (detr + panoptic) or just panoptic head. The last is usefull to optimize forward step, where user would filter boxes before to pass through panoptic head module.

**Updates:**
 * Introduce of `dynamic axes`:  Proposal for the inclusion of `dynamic_axes` in `torch.onnx.export` procedure, as well as on `TRTEngineBuilder`. Inclusion of the optimization profile is required to generate the engine. 
 * Develop of `DynamicHostDeviceMem`, as memory object to allocate dynamic buffer at each forward step.

**Notes:**
* One tested-case of dynamic-axes was implemented in `alonet/detr_panoptic/trt_exporter.py`. Use this script to understand the way of use the news features.
* Deformable architecture was modified to be used by `PanopticHead`. However, deformable + Panoptic has not been tested yet.

### DIFF
--- a/alonet/__init__.py
+++ b/alonet/__init__.py
@@ -1,6 +1,4 @@
 ALONET_ROOT = "/".join(__file__.split("/")[:-1])
-from . import torch2trt
-
 from . import metrics
 from . import common
 from . import detr
@@ -11,3 +9,5 @@ from . import deformable_detr
 from . import callbacks
 
 from . import detr_panoptic
+
+from . import torch2trt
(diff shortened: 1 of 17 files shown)


thibo73800 (commented):
Thanks !

thibo73800 (approved):

PR #150, opened 2022-01-31
Depth.get_view(): fix normalization and add reverse cmap feature

jsalotti:
Bug :
- the cmap was not correctly normalized after clipping

Fix :
- use `vmin`, `vmax` and `clip` arguments of `matplotlib.colors.Normalize`

Feature :
- added `reverse` argument to use colormap in reverse order. Defaults to `True` to stay consistent with previous behavior.


Johansmm (approved):
Done

thibo73800 (approved):

PR #155, opened 2022-03-16
Trt export from onnx

thibo73800:
In this merge request

**Fix and features**
- add `skip_adapt_graph` option to not adapt the graph before to export to TensorRT.
- Fix issue when calling `TRTExecutor()` without engine, its now `TRTExecutor(stream=cuda.Stream())`
- Automatically adapt graph by default: handle clip operations + simplify onnx graph. Is is not mandatory to override this method anymore. This change will not affect the current trt exporter since the `adapt_graph` method is supposed to be override.

### DIFF
--- a/alonet/torch2trt/base_exporter.py
+++ b/alonet/torch2trt/base_exporter.py
@@ -1,6 +1,7 @@
 import io
 import os
 import time
+from tkinter import E
 from typing import Dict, List, Tuple, Union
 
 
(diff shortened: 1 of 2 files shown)


alonet/torch2trt/base_exporter.py
```diff
@@ -147,15 +157,62 @@ def build_torch_model(self):
...
+
+        if check:
+            print("\n[INFO] Simplified ONNX model validated. Graph optimized...")
+            graph = gs.import_onnx(model_simp)
+            graph.toposort()
+            graph.cleanup()
+        else:
+            print("\n[INFO] ONNX model was not validated.")
+
+        if self.use_scope_names:  # Rename nodes to correct profiling
```
jsalotti, line 210: `rename_nodes_` should maybe be called directly in `_torch2onnx` before saving the graph, to keep all the renaming operations in one place.

jsalotti (approved):

Issue #157, opened 2022-03-21
Depth: invert with scale & shift


thibo73800:
Instead of not being sure of the current state of a depth tensor. I propose to encode what the depth tensor is encoding.

```python
depth = aloscene.Depth(...)
depth.is_inverse # False
inverse_depth = depth.encode_inverse()
depth.is_inverse # True
inverse_depth = = inverse_depth.encode_inverse(scale=scale, shift=shift) # encode with the given scane & shift
depth.is_inverse # True
depth.scale 
depth.shift 
depth = inverse_depth.encode_absolute() # encode back as absolute value
depth.is_inverse # False
```

Data-Iab:
when calling `encode_inverse` should we first create a copy of depth then inverse it and keep the first depth as it is, or should the depth be inplaced ? because this line is a little bit confusing.
```python
inverse_depth = depth.encode_inverse()
depth.is_inverse # True
```

PR #159, opened 2022-03-22
add depth absolute/relative encoding

Data-Iab:
Two methods added to **Depth:**
- encode_inverse : invert depth with given scale and shift.
- encode_absolute : undo encode_inverse changes.

One method added to **AugTensor:**
- to_squeezed_numpy: as its name indicates, converts to squeezed numpy.

### DIFF
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -49,11 +49,25 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
     def __init__(self, x, *args, **kwargs):
         super().__init__(x)
 
-    def encode_inverse(
-            self, 
-            scale: float = 1, 
-            shift: float = 0):
-        """Shift and scales the depth values"""
+    def encode_inverse(self, scale=1, shift=0):
+        """Scales and shifts the inverse depth
+        
+        Parameters
+        ----------
+            scale: (: float)
+                Multiplication factor. Default is 1.
+            
+            shift: (: float)
+                Addition intercept. Default is 0.
+        
+        Exemples
+        -------
+        >>> # Create a non inverted depth
+        >>> depth = aloscene.Depth(np.random.normal((1, 20, 20)))
+        >>> inverted_depth = depth.encode_inverse(scale=0.03, shift=1)
+        >>> depth.is_inverse, inverted_depth.is_inverse
+        >>> False, True
+        """
         depth = self
         if depth._is_inverse:
             depth = self.encode_absolute()
@@ -64,7 +78,16 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         return 1 / depth
     
     def encode_absolute(self):
-        """Undo inverse encoding"""
+        """Encodes inverted depth to absolute depth
+        
+        Exemples
+        --------
+        >>> # Create an already inverted depth
+        >>> invert_depth = aloscene.Depth(np.random.normal((1, 20, 20)), scale=0.5, shift=1)
+        >>> depth = depth.encode_absolute(scale=0.03, shift=1)
+        >>> depth.is_inverse, inverted_depth.is_inverse
+        >>> False, True
+        """
         depth = self
         if not depth._is_inverse:
             warnings.warn('depth already encoded in absolute mode')
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -41,9 +41,9 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
             names = ("C", "H", "W")
         tensor = super().__new__(cls, x, *args, names=names, **kwargs)
         tensor.add_child("occlusion", occlusion, align_dim=["B", "T"], mergeable=True)
-        tensor.add_property('_scale', scale)
-        tensor.add_property('_shift', shift)
-        tensor.add_property('_is_inverse', is_inverse)
+        tensor.add_property('scale', scale)
+        tensor.add_property('shift', shift)
+        tensor.add_property('is_inverse', is_inverse)
         return tensor
 
     def __init__(self, x, *args, **kwargs):
@@ -71,9 +71,9 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         depth = self
         if depth._is_inverse:
             depth = self.encode_absolute()
-        depth.add_property('_scale', scale)
-        depth.add_property('_shift', shift)
-        depth.add_property('_is_inverse', True)
+        depth.add_property('scale', scale)
+        depth.add_property('shift', shift)
+        depth.add_property('is_inverse', True)
         depth = depth * scale + shift
         return 1 / depth
     
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -59,7 +59,7 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         super().__init__(x)
 
     def encode_inverse(self):
-        """Undo encode_absolute rtansformation
+        """Undo encode_absolute tansformation
         
         Exemples
         -------
@@ -97,15 +97,15 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         >>> absolute_depth.is_absolute, not_absolute_depth.is_absolute
         >>> True, False
         """
-        depth = self
+        depth, names = self.rename(None), self.names
         if depth.is_absolute:
             raise ExecError('depth already in absolute state, call encode_inverse first')
         depth = depth * scale + shift
-        depth[depth < 1e-8] = 1e-8
+        depth[torch.unsqueeze(depth < 1e-8, dim=0)] = 1e-8
         depth.scale = scale
         depth.shift = shift
         depth.is_absolute = True
-        return 1 / depth
+        return (1 / depth).rename(*names)
 
     def append_occlusion(self, occlusion: Mask, name: str = None):
         """Attach an occlusion mask to the depth tensor.
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -49,7 +49,10 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
     def __init__(self, x, *args, **kwargs):
         super().__init__(x)
 
-    def encode_inverse(self, scale: float, shift: float):
+    def encode_inverse(
+            self, 
+            scale: float = 1, 
+            shift: float = 0):
         """Shift and scales the depth values"""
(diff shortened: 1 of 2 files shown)


aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+            **kwargs):
+        if is_inverse and not (shift and scale):
+            raise AttributeError('inversed depth requires shift and scale arguments')
         if isinstance(x, str):
             x = load_depth(x)
             names = ("C", "H", "W")
         tensor = super().__new__(cls, x, *args, names=names, **kwargs)
         tensor.add_child("occlusion", occlusion, align_dim=["B", "T"], mergeable=True)
-
+        tensor.add_property('_scale', scale)
```
thibo73800, line 44: Any reason to use _scale instead of scale ?

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        tensor.add_property('_shift', shift)
+        tensor.add_property('_is_inverse', is_inverse)
         return tensor
 
     def __init__(self, x, *args, **kwargs):
         super().__init__(x)
 
+    def encode_inverse(self, scale: float, shift: float):
+        """Shift and scales the depth values"""
```
thibo73800, line 53: Here is an example of expected doc string  
```
"""Attach an occlusion mask to the depth tensor.
  Parameters
  ----------
  occlusion: Mask
      Occlusion mask to attach to the Frame
  name: str
      If none, the occlusion mask will be attached without name (if possible). Otherwise if no other unnamed
      occlusion mask are attached to the frame, the mask will be added to the set of mask.
  """
```

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
-
+        tensor.add_property('_scale', scale)
+        tensor.add_property('_shift', shift)
+        tensor.add_property('_is_inverse', is_inverse)
         return tensor
 
     def __init__(self, x, *args, **kwargs):
         super().__init__(x)
 
+    def encode_inverse(self, scale: float, shift: float):
```
thibo73800, line 52: I think that scale & shift should be optional when doing the inverse. By default, encode inverse should only do 1/d

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        """Shift and scales the depth values"""
+        depth = self.detach().cpu()
```
thibo73800, line 54: detach or cpu are not required here.

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        if depth._is_inverse:
+            depth = self.encode_absolute()
+        depth.add_property('_scale', scale)
```
thibo73800, line 57: At this point, the scale property is already added to the augmented tensor.  You only need to set the value you want. depth.scale = scale.

aloscene/depth.py
thibo73800: scale=1 and shift=0 by default

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        depth.add_property('_shift', shift)
+        depth.add_property('_is_inverse', True)
+        depth = depth * scale + shift
+        return 1 / depth
+    
+    def encode_absolute(self):
+        """Undo inverse encoding"""
```
thibo73800, line 64: same for the docstring

aloscene/depth.py
```diff
@@ -22,18 +24,55 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        depth = self.detach().cpu()
```
thibo73800, line 65: same for detach & cpu.

aloscene/tensors/augmented_tensor.py
```diff
@@ -699,6 +699,10 @@ def as_tensor(self):
         n_tensor.__class__ = torch.Tensor
         n_tensor.rename_(None)
         return n_tensor
+    
+    def to_squeezed_numpy(self):
```
thibo73800, line 703: When we think about it, squeeze could be application specific and/or could be made after in numpy
```
np.squeeze(tensor.numpy(), axis=0)
tensor.numpy()[0] # or like this
```
I think this method could be usefull to make it quicker to export to numpy without having to detach & pass to the cpu each time. 

Also the type might be something we want to be set as parameter. 

Finally, its best to first detach, then pass to the cpu (otherwise you copy all the computation graph to the cpu). 

Here is my proposition
- We already have a method named `as_tensor`. Following the same format this method could be named `as_numpy`
- This is different from `tensor.numpy()` that assume you're already on the cpu & detach from graph.
- `as_numpy` will detach, move the tensor on the cpu & finally convert to the expected type.

```
tensor.as_numpy()
tensor.as_numpy(np.float16)
```

Let me know if it make sense to you or if you have an other proposition.

thibo73800 (changes requested):

aloscene/depth.py
Data-Iab: `depth = Depth(... is_inverse=True, scale=1, shift=1) # depth already inverted`
`depth = depth.encode_inverse(scale=2, shift=4) # re-invert with new values`

we need to override the scale and shift once we call the encode_inverse method.

thibo73800: Yes, my point if that `add_property` is used to create the property in the first place. Once the property is created, you don't need to create the property again. You can simply set the value depth.scale = scale

aloscene/tensors/augmented_tensor.py
```diff
@@ -699,6 +699,11 @@ def as_tensor(self):
         n_tensor.__class__ = torch.Tensor
         n_tensor.rename_(None)
         return n_tensor
+    
+    def as_numpy(self, dtype=np.float16):
+        """Returns squeezed numpy array"""
```
thibo73800, line 704: This is obvious, but missing docstring for dtype.

thibo73800 (changes requested):
Once this last comment is solved + the pr description updated, we're gonna be ready to merge. I'll also ask @jsalotti to review since it could happen to use theses methods soon.

jsalotti (commented):
- The transformations are not consistent. Calling `encode_absolute` after `encode_inverse` will not give you the original value. This is due to the order in which you apply substraction and division.
- `encode_inverse` does not seem consistent with the convention of [this paper](https://arxiv.org/pdf/1907.01341.pdf). They apply scale and shift directly on idepth, whereas you apply it on depth before inversion. Is it on purpose ?
- in `encode_inverse`, it seems that the depth is modified in place instead of creating a copy to return an inverse depth

Data-Iab:
- You're right, it's fixed now
- the `encode_inverse` is nothing but [this](https://github.com/isl-org/DPT/blob/f43ef9e08d70a752195028a51be5e1aff227b913/dpt/models.py#L117) with custom scale and shift.
- [issue](https://github.com/Visual-Behavior/aloception/issues/157) : the `is_inverse` change in "depth" when `encode_inverse` is called.

aloscene/depth.py
```diff
@@ -19,21 +21,90 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        >>> # Create an already inverted depth
+        >>> invert_depth = aloscene.Depth(np.random.normal((1, 20, 20)), scale=0.5, shift=1)
+        >>> depth = depth.encode_absolute(scale=0.03, shift=1)
+        >>> depth.is_inverse, inverted_depth.is_inverse
+        >>> False, True
+        """
+        depth = self
+        if not depth.is_inverse:
+            warnings.warn('depth already encoded in absolute mode')
+            return depth
```
thibo73800, line 100: return depth.clone() so that if the user use depth.encode_absolute() the returned depth will be a new tensor

aloscene/depth.py
```diff
@@ -19,21 +21,90 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
...
+        >>> # Create a non inverted depth
+        >>> depth = aloscene.Depth(np.random.normal((1, 20, 20)))
+        >>> inverted_depth = depth.encode_inverse(scale=0.03, shift=1)
+        >>> depth.is_inverse, inverted_depth.is_inverse
+        >>> False, True
+        """
+        depth = self
+        if depth.is_inverse:
+            depth = self.encode_absolute()
+        depth.scale = scale
```
thibo73800, line 80: `depth.scale = scale`
and `depth.shift` and
`depth.is_inverse` must be set after
`depth = depth * scale + shift`

Otherwise you will change both the original tensor and the new depth tensor. 

The best would be for you to write unit test of your code into the unitest folder to check theses possible issues.

aloscene/tensors/augmented_tensor.py
```diff
@@ -699,6 +699,17 @@ def as_tensor(self):
...
+    def as_numpy(self, dtype=np.float16):
+        """Returns squeezed numpy array
+        
+        Parameters
+        ----------
+            dtype: (: np.dtype)
+                The output dtype. Default np.float16.
+        """
+        tensor = self
+        return np.squeeze(tensor.detach().cpu().numpy()).astype(dtype)
```
thibo73800, line 712: As mentioned before, this method is not supposed to squeeze. squeeze can be done by the user itself after having called as_numpy()

thibo73800 (changes requested):

Data-Iab:
It seems that we had a problem with functions name. Calling DPT with `invert=False` means that predictions are inverted (which is not compatible with the attribute name. [here](https://github.com/isl-org/DPT/blob/f43ef9e08d70a752195028a51be5e1aff227b913/dpt/models.py#L123) invert depth is returned). And so we can't call our function `encode_inverse` because there will be more ambiguity.
Hence, I changed methods names so `encode_absolute` is called to transform depth prediction.
```python
dpt = DPT(invert=False)
depth = dpt(frame)  # inverted predicted depth
depth = depth.encode_absolute(shift=shift, scale=scale) # absolute depth
```
`incode_inverse` is called just to undo `encode_absolute` changes

thibo73800:
@Data-Iab Sounds good. I think there is still a last comment unresolved

thibo73800 (approved):

PR #167, opened 2022-04-14
fix depth absolute/inverse assertion

Data-Iab:
Instantiating Depth does not require any conditions now in term of scale/shift/absolute/not absolute. We let the user deal with the depth state and shift/scale attributes.


thibo73800 (approved):

thibo73800:
@Data-Iab  can you resolve the conflicts before I merged the branch ?

thibo73800 (approved):

PR #170, opened 2022-04-20
Conversion between distance and depth

anhtu293:
> The shortest distance between two points is a straight line. - **Archimedes**

As said Archimedes, knowing the distance (straight line) between camera and a point is as important as knowing planar depth. Therefore, it's convenient to have methods that can do the conversion between them

**What's new ?** 

- Handle negative points in **encode_absolute**: For wide range camera (FoV > 180), it's possible to have points whose planar depth is small than 0 (points behind camera). To keep these points instead of clipping by 0, pass keep_negative=True in argument.
- Depth to distance **as_distance()**: Convert depth to distance. Only pinhole camera and linear equidistant camera are supported at this time.
-  Distance to depth **as_depth()**: Convert distance to depth.  Only pinhole camera and linear equidistant camera are supported at this time.
- Possible to create a tensor of Distance by passing is_distance=True at initialization.
- Support functions in depth_utils.

**Update**
-  Change the term to avoid the confusion: "euclidean depth" for distance and "planar depth" for usual depth.
- **as_distance()** becomes **as_euclidean()**
- **as_depth()** becomes **as_planar()**

> Archimedes's quote now becomes: _The shortest "euclidean depth" between two points is a straight line_.

### DIFF
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -81,7 +81,7 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         >>> (undo_depth == not_absolute_depth).item()
         >>> True
         """
-        if not depth.is_absolute:
+        if not self.is_absolute:
             print('No need to inverse depth, already inversed')
             return self.clone()
         depth = self
@@ -130,7 +130,7 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         >>> absolute_depth.is_absolute, not_absolute_depth.is_absolute
         >>> True, False
         """
-        if depth.is_absolute:
+        if self.is_absolute:
             print('Depth already in absolute value.')
             return self.clone()
         depth, names = self.rename(None), self.names
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -138,11 +138,12 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
         if prior_clamp_min is not None or prior_clamp_max is not None:
             depth = torch.clamp(depth, min=prior_clamp_min, max=prior_clamp_max)
 
-        if not keep_negative:
-            depth[torch.unsqueeze(depth < 1e-8, dim=0)] = 1e-8
-        else:
+        if keep_negative and self.is_planar:
             depth[torch.unsqueeze((depth < 1e-8) & (depth >= 0), dim=0)] = 1e-8
             depth[torch.unsqueeze((depth >= -1e-8) & (depth < 0), dim=0)] = -1e-8
+        else:
+            depth[torch.unsqueeze(depth < 1e-8, dim=0)] = 1e-8
+
         depth.scale = scale
         depth.shift = shift
         depth.is_absolute = True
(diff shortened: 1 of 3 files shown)


aloscene/depth.py
```diff
@@ -132,7 +138,11 @@ def encode_absolute(self, scale=1, shift=0, prior_clamp_min=None, prior_clamp_ma
         if prior_clamp_min is not None or prior_clamp_max is not None:
             depth = torch.clamp(depth, min=prior_clamp_min, max=prior_clamp_max)
 
-        depth[torch.unsqueeze(depth < 1e-8, dim=0)] = 1e-8
+        if not keep_negative:
```
thibo73800, line 141: If depth is_planar False then the encode_absolute should not clip value bellow zero, otherwise if is_planar: the negative values are clip. Thus, keep_negative might not be necessary. It should be automatic based of self.is_planar

aloscene/depth.py
```diff
@@ -274,3 +283,74 @@ def as_disp(
...
+            Distortion coefficient for equidistant model. Only linear distortion supported
+            (sensor_angle=distortion*theta).
+
+        Returns
+        -------
+        aloscene.Depth
+        """
+        planar = self
+        if not planar.is_planar:
+            raise ExecError("Cannot transform to euclidian because this is already a euclidian depth tensor.")
```
thibo73800, line 309: Up to know, on other aug tensor, if the operation is call twice, we simply return self.clone() instead of raising an error

aloscene/depth.py
```diff
@@ -274,3 +283,74 @@ def as_disp(
...
+            used for wide range camera).
+        distortion: float | 1.0
+            Distortion coefficient for equidistant model. Only linear distortion supported
+            (sensor_angle=distortion*theta).
+
+        Returns
+        -------
+        aloscene.Depth
+        """
+        planar = self
```
thibo73800, line 307: This method should at least assert the the depth is encoded with absolute value. 
The best would be to adapt the method to handle inverse depth as well and returned an inverse euclidian depth if needed.

aloscene/depth.py
```diff
@@ -274,3 +283,74 @@ def as_disp(
...
+            err_msg += "the as_disp(camera_intrinsic=...) method."
+            raise Exception(err_msg)
+
+        _, theta = coords2rtheta(camera_intrinsic, planar.HW, distortion, projection)
+        euclidean = planar / (torch.cos(theta) + 1e-8)
+        euclidean.is_planar = False
+        return euclidean
+
+    def as_planar(self, camera_intrinsic: aloscene.CameraIntrinsic = None, projection="pinhole", distortion=1.0):
+        """Create a new planar depth augmented tensor from the euclidean depth between camera to world points with
```
thibo73800, line 324: same comment than above

thibo73800 (changes requested):

anhtu293:
**Update**
- In `encode_absolute()` and `encode_inverse()`, if tensor has already in absolute/inverse value, return `self.clone()` instead of raising an error as before.

thibo73800 (approved):

PR #173, opened 2022-04-27
fixed typo l.179

Dee61298:
FitLoop is a PytorchLightning class and its "should accumulate" method starts with a _, the metric doesn't work otherwise.
PL version : 1.5.10


thibo73800:
Hi @Dee61298 Thanks for the fixe. Can you add as a precision you're pl version ? 
Thanks

thibo73800 (changes requested):
We're currently running 1.4.1 officially. https://github.com/Visual-Behavior/aloception/blob/master/requirements.txt#L9 

Moving to 1.5 might happen later

PR #175, opened 2022-05-10
Merge dataset weights

jsalotti:
Add `weights` parameter to `MergeDataset`.
The samples from a dataset with weight `w` will appear `w` times in the MergeDataset.

### DIFF
--- a/alodataset/merge_dataset.py
+++ b/alodataset/merge_dataset.py
@@ -20,18 +20,32 @@ class MergeDataset(torch.utils.data.Dataset):
         List of datasets
     transform_fn : function
         transformation applied to each sample
+    weights: List[int] | None
+        For N datasets, a list of N integer weights.
+        The samples from a dataset with weight `w` will appear `w` times in the MergeDataset.
     """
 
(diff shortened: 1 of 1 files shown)


alodataset/merge_dataset.py
```diff
@@ -20,18 +20,35 @@ class MergeDataset(torch.utils.data.Dataset):
...
+        if any(type(w) != int for w in weights):
+            raise RuntimeError("weights should be a list of int.")
+        return weights
+
     def _init_indices(self):
         indices = []
         for dset_idx, dset in enumerate(self.datasets):
-            for idx in range(len(dset)):
-                indices.append((dset_idx, idx))
+            for _ in range(self.weights[dset_idx]):
```
thibo73800, line 49: Not sure to understand how the following is doing importance sampling based on the weights.
What I understan id that the default scenario is that you have [1, 1] (two dataset). Let's say instead you have [2, 1] to give more importance to the first dataset. You code will do the same thing no matter the weights value.

thibo73800: Let me know if I missed something

PR #176, opened 2022-05-12
add profiling verbosity 

Data-Iab:
New :star: : 
- `profiling_verbosity` option is added to the TRTEngineBuilder to better inspect the details of each node when calling the `tensorrt.EngineInspector`
- Some quantization related arguments are added to the `BaseTRTExporter`.

### DIFF
--- a/alonet/torch2trt/base_exporter.py
+++ b/alonet/torch2trt/base_exporter.py
@@ -408,6 +408,15 @@ class BaseTRTExporter:
         parser.add_argument("--precision", type=str, default="fp32", help="fp32/fp16/mix, default FP32")
         parser.add_argument("--verbose", action="store_true", help="Helpful when debugging")
         parser.add_argument("--profiling_verbosity", default=0, type=int, help="Helpful when profiling the engine")
+        parser.add_argument("--calibration_batch_size", type=int, default=8, help="Calibration data batch size.")
+        parser.add_argument("--limit_calibration_batches", type=int, default=10, help="Limits number of batches.")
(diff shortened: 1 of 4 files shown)


alonet/torch2trt/base_exporter.py
```diff
@@ -390,6 +403,16 @@ def add_argparse_args(parent_parser):
         parser.add_argument("--batch_size", type=int, default=1, help="Engine batch size, default = 1")
         parser.add_argument("--precision", type=str, default="fp32", help="fp32/fp16/mix, default FP32")
         parser.add_argument("--verbose", action="store_true", help="Helpful when debugging")
+        parser.add_argument("--profiling_verbosity", default=0, type=int, help="Helpful when profiling the engine")
```
thibo73800, line 406: Can you add in the help string the default values ?

alonet/torch2trt/quantization.py
```diff
@@ -81,6 +81,9 @@ def set_default_desc(self,  calib_method="histogram", per_channel_quantization=T
         ## convert Layers to QuantLayers
         quant_modules.initialize()
 
+        ## set fake quantization to True before torch2onnx
+        quant_nn.TensorQuantizer.use_fb_fake_quant = True
```
thibo73800, line 85: Why do you have to do that ?

Data-Iab: This is mandatory, otherwise you can't call `torch.export.onnx` with a quantized model

thibo73800 (approved):

PR #180, opened 2022-05-17
clip low values of depth before conversion to disp

jsalotti:
Modify the behavior of `Disparity.as_disp()` :

- clip low value of depth to 1e-8 to avoid extremely large disparity values

Off-topic : lot of modifications in this pull request come from auto-formatting. We should stress again that everybody shoud use it if we don't want to clutter pull-requests with formatting edits.


thibo73800:
Add me for review once this is not a  draft anymore

thibo73800 (approved):

PR #181, opened 2022-05-18
Fix package prod dependencies

thibo73800:
In this merge request:

- Move `pytorch_quantization` as an optional prod `dependency`


Data-Iab:
We don't use `pytorch_quantization` in the base exporter anymore

PR #182, opened 2022-05-18
remove tracing assertion

Data-Iab:
Quick fix:
- Tracing assertion is removed from the BaseExporter.


thibo73800:
Is it not true that all model with the tracing property must be exported to onnx/tensorRT with tracing=True ?

Data-Iab:
@thibo73800 well it depends on what `tracing` means, in AloDPTDepthSegmentation for example it does not refer to the same thing. For general purposes, we better handle assertions in child classes.

thibo73800 (approved):

PR #188, opened 2022-05-31
Dynamic cropping

anhtu293:
**Dynamic Cropping**
- Possibility to crop an image to smaller fixed size image in the position we want. The crop position can be parsed by argument `center` which can be `float` or `int`.
- If crop is out of image border, an error is triggered.

### DIFF
--- a/alodataset/transforms.py
+++ b/alodataset/transforms.py
@@ -881,8 +881,8 @@ class DynamicCropTransform(AloTransform):
                 Default: (0.5, 0.5)
         """
         if isinstance(center[0], float):
-            center_x = int(frame.W * center[0])
-            center_y = int(frame.H * center[1])
+            center_x = int(frame.W * center[0]) - 1
+            center_y = int(frame.H * center[1]) - 1
         else:
             center_x = center[0]
             center_y = center[1]
--- a/alodataset/transforms.py
+++ b/alodataset/transforms.py
@@ -870,10 +870,10 @@ class DynamicCropTransform(AloTransform):
         super().__init__(*args, **kwargs)
 
     def sample_params(self):
-        return tuple()
+        return (self.crop_size,)
 
-    def set_params(self):
-        pass
+    def set_params(self, size):
+        self.crop_size = size
 
     def apply(self, frame: Frame, center: Union[Tuple[int, int], Tuple[float, float]] = (0.5, 0.5)):
         """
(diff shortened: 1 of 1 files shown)


thibo73800 (changes requested):

alodataset/transforms.py
```diff
@@ -858,3 +858,43 @@ class RandomDownScaleCrop(Compose):
...
+    """ Crop image to target crop size at chosen position.
+    """
+    def __init__(self, crop_size, *args, **kwargs):
+        assert all([isinstance(s, int) for s in crop_size])
+        self.crop_size = crop_size
+
+        super().__init__(*args, **kwargs)
+
+    def sample_params(self):
+        return tuple()
```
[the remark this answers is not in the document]
anhtu293: Yes it's not an issue in my case but I just added also the `sample_params` et `set_params` as in other classes like Rotate and Resize.

alodataset/transforms.py
```diff
@@ -858,3 +858,43 @@ class RandomDownScaleCrop(Compose):
...
+            center_x = center[0]
+            center_y = center[1]
+
+        crop_h, crop_w = self.crop_size[0], self.crop_size[1]
+        left = center_x - crop_w // 2
+        top = center_y - crop_h // 2
+        right = center_x + crop_w // 2
+        bot = center_y + crop_h // 2
+
+        if left < 0 or top < 0 or right > (frame.W - 1) or bot > (frame.H - 1):
```
Data-Iab, line 896: I don't know if that's gonna be useful or not but ValueError raises when W or H is the same as one of the corresponding crop sizes (i.e if we want to keep the same frame's H or W and crop the other dim)
Exemple :

```python
center = [0.5, 0.5]
frame.W = 50
frame.H = 20
crop_size = [20, 20]
```

```python
ValueError: Crop coordinates out of image border. Image size: (50, 20), Crop coordinate (top, left, bot, right): (0, 15, 20, 35)
```

Data-Iab, line 896: I don't know if that's gonna be useful or not but ValueError raises when W or H is the same as one of the corresponding crop sizes (i.e if we want to keep the same frame's H or W and crop the other dim)
Exemple :

```python
center = [0.5, 0.5]
frame.W = 50
frame.H = 20
crop_size = [20, 20]
```

```python
ValueError: Crop coordinates out of image border. Image size: (50, 20), Crop coordinate (top, left, bot, right): (0, 15, 20, 35)
```

anhtu293: You're right. It's an error in my coordinate calculation. fix commit will be pushed quickly

alodataset/transforms.py
```diff
@@ -881,8 +881,8 @@ def apply(self, frame: Frame, center: Union[Tuple[int, int], Tuple[float, float]
                 Default: (0.5, 0.5)
         """
         if isinstance(center[0], float):
-            center_x = int(frame.W * center[0])
-            center_y = int(frame.H * center[1])
+            center_x = int(frame.W * center[0]) - 1
```
Data-Iab, line 884: I don't think that shifting the center is good choice. with the same example above you still have the same problem.
```python
ValueError: Crop coordinates out of image border. Image size: (50, 20), Crop coordinate (top, left, bot, right): (-1, 14, 19, 34)
```

thibo73800 (approved):

thibo73800:
Will merge once @Data-Iab approved

Data-Iab (approved):
I think it's all good

PR #193, opened 2022-06-20
add dataset merging order

Data-Iab:
New :star: 
- `MergeDataset` comes with `order` argument to order samples from merged datasets.
  `order=[2, 1]` will force the iterator to return 2 samples from the first dataset then 1 sample from the second one...

### DIFF
--- a/alodataset/merge_dataset.py
+++ b/alodataset/merge_dataset.py
@@ -1,4 +1,5 @@
 import torch
+import random
 import numpy as np
 
 from alodataset.base_dataset import rename_data_to_none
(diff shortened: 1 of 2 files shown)


alodataset/merge_dataset.py
```diff
@@ -18,18 +20,48 @@ class MergeDataset(torch.utils.data.Dataset):
...
     """
-
-    def __init__(self, datasets, transform_fn=None, weights=None):
+    def __init__(
+            self,
+            datasets,
+            order=None,
+            shuffle=False,
+            lim_samples=None,
+            transform_fn=None,
```
thibo73800, line 41: Is it still possible to use `weights` ? This is still in the doc but not use in the code anymore. Also it seem to me that weights was a good parameter to decide on how to sample from each dataset.

thibo73800 (changes requested):

thibo73800 (commented):
What happen when a datasets is smaller than an other but two time more weighted.
Say `datasets=[d1, d2]` with `d1` two times smaller than `d2` but the weights are `[2, 1]` ?

Data-Iab:
@thibo73800 we respect the order until one of the datasets has come to an end. In this case we can find all ds1 samples, but only part of ds2, more precisely, the number of samples from ds2 is `len(ds1) / ds1_weight * ds2_weight = len(ds1) / 2`

thibo73800:
What is the status of this PR (when you need me to review it again you have a button for that)

Issue #160, opened 2022-03-24
Dimension mismatch when masking AugmentedTensor


Data-Iab:
It seems like AugmentedTensor is unable to mask values:

```python
>>> depth = Depth(torch.ones((1, 7, 7)), names=('B', 'H', 'W'))
>>> (depth == 1).shape  # torch.Size([1, 7, 7])
>>> depth.shape  # torch.Size([1, 7, 7])
>>> depth[depth == 1]  #  Error here
```
outputs:
```
  File "/aloception/aloscene/tensors/augmented_tensor.py", line 305, in __getitem__
    tensor = torch.Tensor.__getitem__(self.rename(None), idx).reset_names()
  File "/aloception/aloscene/tensors/augmented_tensor.py", line 566, in __torch_function__
    tensor = super().__torch_function__(func, types, args, kwargs)
  File "/home/ubuntu/anaconda3/envs/pytorch_1100_2/lib/python3.8/site-packages/torch/_tensor.py", line 1051, in __torch_function__
    ret = func(*args, **kwargs)
IndexError: The shape of the mask [7, 7] at index 0 does not match the shape of the indexed tensor [1, 7, 7] at index 0

```

PR #194, opened 2022-06-21
fix masking dimension mismatch

Data-Iab:
Fix :wrench: : AugmentedTensor can be masked without any need to expand the mask dimensions.

### DIFF
--- a/aloscene/tensors/augmented_tensor.py
+++ b/aloscene/tensors/augmented_tensor.py
@@ -302,7 +302,9 @@ class AugmentedTensor(torch.Tensor):
                 )
 
         if isinstance(idx, torch.Tensor):
-            tensor = torch.Tensor.__getitem__(self.rename(None), idx).reset_names()
+            if not all(e in [0, 1] for e in idx.unique().int()): 
+                raise IndexError(f"Unvalid mask. Expected mask elements to be in [0, 1, True, False]")
+            tensor = self * idx
         else:
             tensor = torch.Tensor.__getitem__(self, idx)
 
--- a/aloscene/tensors/augmented_tensor.py
+++ b/aloscene/tensors/augmented_tensor.py
@@ -302,8 +302,9 @@ class AugmentedTensor(torch.Tensor):
                 )
 
         if isinstance(idx, torch.Tensor):
-            if not all(e in [0, 1] for e in idx.unique().int()): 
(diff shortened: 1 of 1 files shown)


Data-Iab:
Potential fix for this [issue](https://github.com/Visual-Behavior/aloception/issues/160)

aloscene/tensors/augmented_tensor.py
```diff
@@ -302,7 +302,9 @@ def __getitem__(self, idx):
                 )
 
         if isinstance(idx, torch.Tensor):
-            tensor = torch.Tensor.__getitem__(self.rename(None), idx).reset_names()
+            if not all(e in [0, 1] for e in idx.unique().int()):
```
thibo73800, line 305: I'm just worry about this line since it make filtering data slower by getting the `unique()` of the `idx`. What what the purpose of raising an Exception here ? Just getting a better error message ? If so, you can try catch the tensor & then check the possible source of error.

thibo73800 (changes requested):

aloscene/tensors/augmented_tensor.py
Data-Iab: @thibo73800 It is very important to raise the exception to make sure the user is not unintentionally passing something else rather than a filter. For the time problem you're right, here is a potential solution that can replace this line:
`if not torch.equal(idx ** 3, idx)` as the only solution for `x ** 3 = x` equation is either 0 or 1, it ensures that they are the only values `idx` have.
Concerning running time, you can take a look at the results of the following code:
```python

m = torch.ones((1, 500, 500, 3))

t0 = time.time()
for _ in range(1000):
  if not torch.equal(m ** 3, m):
    pass


t1 = time.time()
print("runtime with torch.equal", t1 - t0)

t0 = time.time()
for _ in range(1000):
  if not all(e in [0, 1] for e in m.unique().int()):
    pass


t1 = time.time()
print("runtime with with previous solution", t1 - t0)
```
```
runtime with torch.equal 1.6653542518615723
runtime with with previous solution 17.1923611164093
```

thibo73800: Ok, what happen if `the user is unintentionally passing something else rather than a filter. ` ? The code will fail ?

thibo73800:
What is the status of this PR ?

thibo73800 (approved):

PR #196, opened 2022-07-06
Lvis Dataset + Coco Update + minor fix

thibo73800:
### New small features
- `CocoDetectionDataset` can now use a given `ann_file` when loaded
- `CocoPanopticDataset` can now use `ignore_classes` to ignore some classed when loading the panoptic anns
- In `DetrCriterion` interpolation is an option that can be changed with `upscale_interpolate`

### New dataset
- Lvis Dataset based on `CocoDetectionDataset` with a different ann file

### Bug fix

- `GLOBAL_COLOR_SET_CLASS` will automaticly adjust its size for giving random color for a given object class

### DIFF
--- a/alodataset/coco_panoptic_dataset.py
+++ b/alodataset/coco_panoptic_dataset.py
@@ -45,6 +45,8 @@ class CocoPanopticDataset(BaseDataset, SplitMixin):
         Include masks labels in the output, by default True
     classes : list, optional
         List of classes to be filtered in the annotation reading process, by default None
+    ignore_classes: list, optional
+        List of classed to ignore in the annotation reading process, by default None.
     fix_classes_len : int, optional
         Fix to a specific number the number of classes, filling the rest with "N/A" value.
         Use when the number of model outputs does not match with the number of classes in the dataset, by default 250
@@ -87,7 +89,6 @@ class CocoPanopticDataset(BaseDataset, SplitMixin):
         self.label_names, self.label_types, self.label_types_names = None, None, None
         self.items = self._get_sequences()
 
-        #print('label_names', self.label_names)
         if classes is not None and ignore_classes is not None:
             raise Exception("Can't considere both classes & ignore_classes at the same time.")
         if ignore_classes is not None:
--- a/alodataset/coco_panoptic_dataset.py
+++ b/alodataset/coco_panoptic_dataset.py
@@ -92,7 +92,7 @@ class CocoPanopticDataset(BaseDataset, SplitMixin):
         if classes is not None and ignore_classes is not None:
             raise Exception("Can't considere both classes & ignore_classes at the same time.")
         if ignore_classes is not None:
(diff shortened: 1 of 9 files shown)


alonet/detr_panoptic/criterion.py
```diff
@@ -133,14 +134,23 @@ def loss_masks(self, outputs: dict, frames: aloscene.Frame, indices: list, num_b
...
         # Masks resize
         fr_shape = frames.shape[-2:]  # (H,W)
-        outputs_masks = F.interpolate(outputs["pred_masks"], size=fr_shape, mode="bilinear", align_corners=False)
+
+        if self.upscale_interpolate:
+            outputs_masks = F.interpolate(outputs["pred_masks"], size=fr_shape, mode="bilinear", align_corners=False)
+        else:
+            outputs_masks = outputs["pred_masks"]
+            o_shape = outputs_masks.shape[-2:]
+            pass
```
Data-Iab, line 143: no operation

alodataset/lvis_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+            if "train2017" not in path:
+             filtered_item.append(item)
+        assert type(self.items) == type(filtered_item)
+        self.items = filtered_item
+
+
+
+    def load_anns(self, index):
+        """
+        Args:
```
Data-Iab, line 51: you can stick with the same docstring form as in getitem
```python
"""
Parameters
----------
returns
-------
"""
```

alodataset/lvis_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+
+# import detr.datasets.transforms as T
+
+from torchvision.datasets.coco import CocoDetection
+from PIL import Image
+from alodataset import BaseDataset, CocoDetectionDataset
+from aloscene import BoundingBoxes2D, Frame, Labels, Mask
+
+
+
```
Data-Iab, line 21: Could be better if you can adjust the number of blank lines in this file

alodataset/coco_panoptic_dataset.py
```diff
@@ -86,6 +89,11 @@ def __init__(
         self.label_names, self.label_types, self.label_types_names = None, None, None
         self.items = self._get_sequences()
 
+        if classes is not None and ignore_classes is not None:
+            raise Exception("Can't considere both classes & ignore_classes at the same time.")
+        if ignore_classes is not None:
+            classes = [l for l in self.label_names if l not in ignore_classes and l != 'N/A']
```
Data-Iab, line 95: I know the two lists are short but we can have faster execution with sets:
```python
ign_cls = set(ignore_classes)
ign_cls.add("N/A")
classes = [l for l in self.label_names if l not in ign_cls]
```

Data-Iab (changes requested):

Data-Iab (approved):

PR #201, opened 2022-07-15
allow user to complety specify the grid view

jsalotti:
Add `from_list_of_list` argument to `Renderer.get_view()`. 

This allows the user to give a list of list of views, defining the 2D grid of views.
All views keep their original shape, and are top-left aligned in their rows.

# Example
```
# Create three gray frames, display them on two rows (2 on first rows, 1 on 2nd row)
import numpy as np
import aloscene
arrays = [np.full((3, 600, 650), 100), np.full((3, 500, 500), 50), np.full((3, 500, 800), 200)]
frames = [aloscene.Frame(arr) for arr in arrays]
views = [[frames[0].get_view(), frames[1].get_view()], [frames[2].get_view()]]
aloscene.render(views, renderer="matplotlib")
```

### DIFF
--- a/aloscene/renderer/renderer.py
+++ b/aloscene/renderer/renderer.py
@@ -178,6 +178,18 @@ class Renderer(object):
 
     @staticmethod
     def add_title(array, start, title):
+        """
+        Add a box with view title
+
+        Parameters
+        ----------
+        array : np.ndarray
+            grid view array (will be modified inplace)
+        start : tuple
+            top-left corner of title box
+        title : str
+            title of the view
+        """
         if title is None:
             return
         else:
--- a/aloscene/renderer/renderer.py
+++ b/aloscene/renderer/renderer.py
@@ -142,8 +142,8 @@ class Renderer(object):
         plt.imshow(view)
         plt.show()
 
-    @staticmethod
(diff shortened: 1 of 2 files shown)


aloscene/renderer/renderer.py
```diff
@@ -220,8 +251,10 @@ def render(
             raise ValueError("The renderer must be one of the following:{}".format(self.renderer_to_fn.keys()))
         if skip_views and record_file is None:
             raise ValueError("When skip_views is desired, a record_file must be provided.")
-
-        view = self.get_grid_view(views, cell_grid_size=cell_grid_size, grid_size=grid_size)
+        if from_list_of_list:
+            view = self.get_user_defined_grid_view(views)
```
Data-Iab, line 255: Can't we replace it with `isinstance(views[0], list)` instead ? it does not seem name-related to what it does

Data-Iab (approved):
Tested.

jsalotti:
@Data-Iab I added the changes you suggested. The branch is ready to be merged.

thibo73800:
@jsalotti  Can you add in the description a small example of how one could use it ? I'll merge after

jsalotti:
> @jsalotti Can you add in the description a small example of how one could use it. So that it can be described in the next release ? I'll merge after

@thibo73800 done.

PR #202, opened 2022-07-15
allow for non integer principal point coordinates

jsalotti:
Removed a constraint that forced image dimensions to be even in order to have integer principal point in camera intrinsic matrix.

### DIFF
--- a/aloscene/camera_calib.py
+++ b/aloscene/camera_calib.py
@@ -64,10 +64,6 @@ class CameraIntrinsic(AugmentedTensor):
             x[0][1] = skew if skew is not None else 0
 
             if principal_point is None and plane_size is not None:
-                if plane_size[0] % 2 != 0 or plane_size[1] % 2 != 0:
-                    err = "The principal points (center of the plane) can't be infer automaticly."
-                    err += " The latter must be given when creating the CameraIntrinsic tensor."
-                    raise Exception(err)
                 principal_point = (plane_size[0] / 2, plane_size[1] / 2)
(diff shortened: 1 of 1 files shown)


aloscene/camera_calib.py
```diff
@@ -64,10 +64,6 @@ def __new__(
             x[0][1] = skew if skew is not None else 0
 
             if principal_point is None and plane_size is not None:
-                if plane_size[0] % 2 != 0 or plane_size[1] % 2 != 0:
-                    err = "The principal points (center of the plane) can't be infer automaticly."
-                    err += " The latter must be given when creating the CameraIntrinsic tensor."
-                    raise Exception(err)
                 principal_point = (plane_size[0] / 2, plane_size[1] / 2)
```
Data-Iab, line 67: Isn't it necessary for the principal point to be integer?

jsalotti: I do not see where it is necessary. If somewhere in the code, maybe the check should be done only there. 
Currently, if you resize a `SpatialAugmentedTensor` with a `cam_intrinsic` labels, the principal points can already become float, and it seems to be working as expected.

anhtu293: I think principal point should be in float so that it could work with real calibrated camera too

Data-Iab (approved):

aloscene/camera_calib.py
thibo73800: Make sense

thibo73800 (approved):

PR #205, opened 2022-07-22
GitHub actions

Ardorax:
New:
- Github action who automatically launch unit test when there is a commit or pull request in master branch.

Question:
- [x] Which version of the packages should we use ?
- [ ] Should we check more branch ?

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - 'master'
-      - 'github_actions'
   pull_request:
     branches:
       - 'master'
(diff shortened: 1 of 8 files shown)


thibo73800 (changes requested):
All tests do not seem to pass. Is it still a WIP or do you need help to find out why tests do not pass ?

Ardorax:
I will change the version of pytorch to 10.2. If there is still some problem. I will need some help.

thibo73800:
@Data-Iab unit test passed and can now be tested automatically (thanks @Ardorax). But, to make it work I had to remove the changes of one of your last pull request (https://github.com/Visual-Behavior/aloception/pull/194/files). 

When using your changes, it seem that aug tensor are not slice properly when using a bool tensor. It was one of the reason unit tests was not passing anymore.

thibo73800 (approved):

PR #208, opened 2022-08-04
Scene flow in frame

Ardorax:
New:

New method 'append_scene_flow' in frame class.

### DIFF
--- a/aloscene/frame.py
+++ b/aloscene/frame.py
@@ -126,6 +126,7 @@ class Frame(aloscene.tensors.SpatialAugmentedTensor):
         tensor.add_child("segmentation", segmentation, align_dim=["B", "T"], mergeable=False)
         tensor.add_child("labels", labels, align_dim=["B", "T"], mergeable=True)
         tensor.add_child("pose", labels, align_dim=["B", "T"], mergeable=True)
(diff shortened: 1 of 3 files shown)


aloscene/frame.py
```diff
@@ -318,6 +330,19 @@ def append_pose(self, pose: Pose, name: str = None):
         """
         self._append_child("pose", pose, name)
 
+    def append_scene_flow(self, scene_flow: SceneFlow, name: str = None):
```
Data-Iab, line 333: You'll need to add the scene flow label in `__new__`

aloscene/frame.py
```diff
@@ -318,6 +330,19 @@ def append_pose(self, pose: Pose, name: str = None):
...
+        """Attach a scene flow to the frame.
+
+        Parameters
+        ----------
+        scene_flow : :mod:`SceneFlow <aloscene.SceneFlow>`
+            Depth to attach to the Frame
+        name : str
+            If none, the scene flow will be attached without name (if possible). Otherwise if no other unnamed
+            scene flow is attached to the frame, the scene flow will be added to the set of scene flows.
+        """
```
Data-Iab, line 343: Can you add an example in the docstring

Data-Iab (changes requested):

Data-Iab (approved):

PR #213, opened 2022-08-17
Kitti datasets

Ardorax:
Kitti Dataset (Stereo, Flow, Scene Flow, Depth, Odometry, Object, Tracking, Road, Semantics)

* ***Kitti Depth*** : How to use Kitti Depth
```python
date = "2011_09_26"
idsOfDrives = [
    "0001",  # sample from training subset
    "0002",  # sample from validation subset
]
custom_drives = {date: idsOfDrives}
kitti_ds = KittiDepth(
    subset="all",
    return_depth=True,
    custom_drives=custom_drives,
)

for f, frames in enumerate(kitti_ds.train_loader(batch_size=2)):
    frames = Frame.batch_list(frames)
```

* ***Kitti Semantic*** : The semantic class
```python
dataset = KittiSemanticDataset()
obj = dataset.getitem(0)
obj.get_view().render()
```

* ***How to use the remaining task of the dataset*** : Dataset's class list : `KittiStereoFlow2012`, `KittiStereoFlowSFlow2015`, `KittiOdometryDataset`, `KittiObjectDataset`, `KittiTrackingDataset`, `KittiRoadDataset`
```python
dataset = DATASET_CLASS(right_frame=False)
obj = dataset.getitem(0)
obj["left"].get_view().render()
```

* ***Scene Flow: dimensions*** : Error with shape of occlusion mask.

_____
This pull request includes

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/alodataset/kitti_depth.py
+++ b/alodataset/kitti_depth.py
@@ -1,0 +1,240 @@
+import os
+import glob
+import warnings
+import numpy as np
+from PIL import Image
+from typing import List, Dict, Union
+from aloscene import Frame, Depth
+from alodataset import BaseDataset
+
+
+class KittiDepth(BaseDataset):
+    """
+    Parameters
+    ----------
+        subset: either 'train', 'val' or 'all'. If all, both val and train subset are considered.
+
+        return_depth: Either to include depth in the output or not.
+
+        custom_drives: Dictionnary (Keys (:str) registration date of the drive ('YYYY_MM_DD'). Values: (:List[Union[str, Dict]])
+                       either ids of drives or dictionnaries of drives of ids and specific files.
+                       Exemples: custom_drives = {'2011_09_28': ['0222', '0186']} will load all images from
+                                 2011_09_28_drive_0222_sync and 2011_09_28_drive_0186_sync.
+                                 custom_drives = {'2011_09_28': {'0222': ['0000000015', '0000000010']}} will load the two images.
+                       Ids should match the drives dates, otherwise the drive folder will be ignored.
+                       If None, all images from specified subset are loaded.
+
+        name: Key of database name in :attr:`alodataset_config.json` file, by default *kitti*.
+
+        main_folder: The main folder of input images, could be image_00, image_01, image_02 or image_03.
+
+        sample: Download (or not) a dataset sample from internet and replace the default dataset_dir.
+
+
+    Attributes
+    ----------
+        drives: List of paths to kitti drive folders.
+
+        drives_folders: List of paths to drives' parent forlders.
+
+
+    Exemples
+    --------
+        >>> # get all the dataset
+        >>> kitti_ds = KittiBaseDataset(
+                                subset='all',
+                                return_depth=True,
+                                )
+        >>> # get specific files from both val and train
+        >>> idsOfDrives = ['0001',  # drive from the training subset
+                           '0002',  # drive from the validation subset
+                          ]
+        >>> date = '2011_09_26'
+        >>> custom_drives = {date: idsOfDrives}
+        >>> kitti_ds = KittiBaseDataset(
+                                return_depth=True,
+                                custom_drives=custom_drives,
+                                )
+    """
+
+    VALID_SUBSETS = ["train", "val", "all"]
+    VALID_MAIN_FOLDERS = ["image_00", "image_01", "image_02", "image_03"]
+
+    def __init__(
+        self,
+        subset: str = "all",
+        return_depth: bool = True,
+        custom_drives: Union[Dict[str, List[str]], None] = None,
+        name: str = "kitti",
+        main_folder: str = "image_02",
+        sample: bool = False,
+        **kwargs,
+    ):
+        if sample:
+            raise NotImplementedError("Download option is not yet available for kitti.")
+        super(KittiDepth, self).__init__(name=name, **kwargs)
+
+        assert main_folder in self.VALID_MAIN_FOLDERS, f"main_folder should be in {self.VALID_MAIN_FOLDERS}"
+        assert subset in self.VALID_SUBSETS, f"subset should be one of {self.VALID_SUBSETS}"
+
+        self.main_folder = main_folder
+        self.return_depth = return_depth
+        self.drives = list()
+        self.drives_folders = list()
+        self.subset = subset
+        self.custom_drives = custom_drives
+        if self.subset == "all":
+            self.subset = self.VALID_SUBSETS[:2]
+        else:
+            self.subset = [self.subset]
+
+        for ss in self.subset:
+            self.drives_folders.append(os.path.join(self.dataset_dir, "images", ss))
+
+        if custom_drives is None:
+            for dfolder in self.drives_folders:
+                self.drives += glob.glob(os.path.join(dfolder, "*"))
+
+        else:
+            for date in custom_drives.keys():
+                for drive_id in custom_drives[date]:
+                    drive_exits = False
+                    if isinstance(drive_id, str):
+                        drive_name = self.get_drive_name(date, drive_id)
+                    elif isinstance(drive_id, dict):
+                        drive_name = self.get_drive_name(date, list(drive_id.keys())[0])
+                    else:
+                        raise AttributeError("invalid custom drive format")
+
+                    for dfolder in self.drives_folders:
+                        path_to_drive = os.path.join(dfolder, drive_name)
+                        if os.path.isdir(path_to_drive):
+                            drive_exits = True
+                            self.drives.append(path_to_drive)
+                            # break if path exists in the validation subset
+                            break
+                    if not drive_exits:
+                        print(
+                            f"Warning: no drive named {drive_name}. Drive id and date do not match or drive is not in the subset"
+                        )
+
+        for drive in self.drives:
+            if self.custom_drives:
+                basename = os.path.basename(drive)
+                date, drive_id = self.date_id_from_drive(basename)
+                if isinstance(self.custom_drives[date], list):
+                    self.items += self._get_paths_from_subdirs(drive)
+                if isinstance(self.custom_drives[date], dict):
+                    self.items += self._get_paths_from_subdirs(drive, self.custom_drives[date][drive_id])
+
+            else:
+                self.items += self._get_paths_from_subdirs(drive)
+
+        # making sure corresponding depth files exist
+        if self.return_depth:
+            replace_items = list()
+            for item in self.items:
+                depth_path = self.get_corresponding_depth(item)
+                if os.path.exists(depth_path):
+                    replace_items.append(item)
+            lost_items = len(self.items) - len(replace_items)
+            if lost_items > 0:
+                print(f"ground truth images not found for {lost_items} images")
+                self.items = replace_items
+
+        if self.custom_drives and not self.items:
+            raise AttributeError("Could not read any images from the given custom drives")
+
+    @classmethod
+    def from_yaml(cls, path: str, **kwargs):
+        """pass custom drives from a .yaml file"""
+        import yaml
+
+        with open(path, "r") as f:
+            custom_drives = yaml.load(f, Loader=yaml.FullLoader)
+        return cls(custom_drives=custom_drives, **kwargs)
+
+    @staticmethod
+    def depth_from_path(path: str) -> Depth:
+        """reads depth from the given path"""
+        depth = np.asarray(Image.open(path), dtype=np.int16)
+        return Depth(np.expand_dims(depth, 0))
+
+    def getitem(self, idx):
+        """Get the :mod:`Frame <aloscene.frame>` corresponds to *idx* index
+
+        Parameters
+        ----------
+        idx : int
+            Index of the frame to be returned
+
+        Returns
+        -------
+        :mod:`Frame <aloscene.frame>`
+            Frame with its corresponding depth
+        """
+        if self.sample:
+            return BaseDataset.__getitem__(self, idx)
+
+        image_path = self.items[idx]
+        frame = Frame(image_path)
+
+        if self.return_depth:
+            depth_path = self.get_corresponding_depth(image_path)
+            depth = self.depth_from_path(depth_path)
+            frame.append_depth(depth)
+
+        return frame
+
+    def get_corresponding_depth(self, path: str):
+        """
+        Args:
+            path to drive
+        Returns:
+            corresponding path to depth groud truth image.
+        """
+        fixed_dir = os.path.join("proj_depth", "groundtruth")
+        filename = os.path.basename(path)
+        driveAndSubset = os.path.join(*path.split(os.sep)[-5:-3])
+        return os.path.join(self.dataset_dir, "depth", driveAndSubset, fixed_dir, self.main_folder, filename)
+
+    def _get_paths_from_subdirs(
+        self, drive_path: str, filenames: Union[List[str], None] = None, extension: str = "png"
+    ) -> List[str]:
+        """return paths to images in drive_path/self.main_folder/data with the specified extension"""
+        assert extension in ["png", "jpg"], "unvalid extension: should be 'png' | 'jpg'"
+        main_path = os.path.join(drive_path, self.main_folder, "data")
+
+        if not os.path.exists(main_path):
+            drive = os.path.basename(drive_path)
+            print(f"main images folder {self.main_folder} not found for drive {drive}")
+        if filenames:
+            return [os.path.join(main_path, f"{filename}.{extension}") for filename in filenames]
+        else:
+            return sorted(glob.glob(os.path.join(main_path, f"*.{extension}")))
+
+    @staticmethod
+    def get_drive_name(date: str, drive_id: str) -> str:
+        """return the drive file name with the corresponding date and drive id"""
+        return date + "_drive_" + drive_id + "_sync"
+
+    @staticmethod
+    def date_id_from_drive(basename):
+        return basename[:10], basename[17:21]
+
+
+if __name__ == "__main__":
+    date = "2011_09_26"
+    idsOfDrives = [
+        "0001",  # sample from training subset
+        "0002",  # sample from validation subset
+    ]
+    custom_drives = {date: idsOfDrives}
+    kitti_ds = KittiDepth(
+        subset="all",
+        return_depth=True,
+        custom_drives=custom_drives,
+    )
+
+    for f, frames in enumerate(kitti_ds.train_loader(batch_size=2)):
+        frames = Frame.batch_list(frames)
--- a/alodataset/kitti_depth.py
+++ b/alodataset/kitti_depth.py
@@ -10,6 +10,7 @@ from alodataset import BaseDataset
 
 class KittiDepth(BaseDataset):
     """
+    Depth task from KITTI dataset.
     Parameters
     ----------
         subset: either 'train', 'val' or 'all'. If all, both val and train subset are considered.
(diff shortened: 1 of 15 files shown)


thibo73800:
You can follow this format for your pull request. (should be automatic next time)
https://github.com/Visual-Behavior/aloception/blob/master/.github/pull_request_template.md

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,273 @@
...
+
+        self.dataset_dir = os.path.join(self.dataset_dir, "dataset")
+
+        self.sequences = self._load_sequences(sequences)
+
+        self.items: Dict[Any, Any] = {"seq_params": {}}
+
+        for seq in self.sequences:
+            # Conpute sequence parameters once and store them in the items dictionary.
+            self.items["seq_params"][seq] = {}
```
jsalotti, line 51: Create `self.seq_params` instead of storing it in `self.items`. 
`self.items` does not store metadata for a whole sequence of the dataset, but only for a sample of data. The keys are always of type int and represent the position of the sample in the final dataset.

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,273 @@
...
+                        "temporal_sequence": temporal_seq,
+                    }
+                    for idx, temporal_seq in enumerate(temporal_sequences)
+                }
+            )
+
+    def __len__(self):
+        if self.sample:
+            return len(self.items)
+        return len(self.times)
```
jsalotti, line 91: no special case needed w.r.t. `self.sample`: in both cases, return `len(self.items) `

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,273 @@
...
+        -------
+        Dict[str, Frame]
+            Dictionary with the loaded frame and pose.
+        """
+
+        if self.sample:
+            return BaseDataset.__getitem__(self, idx)
+
+        item = self.items[idx]
+        print("item", item)
```
jsalotti, line 134: remove `print` statement when you are not debugging anymore.

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,273 @@
...
+        self.skip = skip
+        self.sequence_skip = sequence_skip
+
+        if self.sample:
+            return
+
+        self.dataset_dir = os.path.join(self.dataset_dir, "dataset")
+
+        self.sequences = self._load_sequences(sequences)
+
```
jsalotti, line 46: you should check that the selected sequences are all compatible with `self.split`

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,273 @@
...
+    def __len__(self):
+        if self.sample:
+            return len(self.items)
+        return len(self.times)
+
+    def _load_sequences(self, sequences) -> List[str]:
+        loaded_sequences = []
+
+        if sequences is None:
```
jsalotti, line 96: if None, load all sequences compatible with the value of `self.split`

jsalotti (changes requested):
I added a review for `kitti_odometry`. I will soon add comment for stereo datasets.

alodataset/kitti_stereo_flow2012.py
```diff
@@ -0,0 +1,244 @@
...
+        if self.sample:
+            return
+
+        assert sequence_start <= sequence_end, "sequence_start should be less than sequence_end"
+
+        if "disp_noc" in load or "disp_occ" in load or "disp_refl_noc" in load or "disp_refl_occ" in load:
+            assert sequence_start <= 11 and sequence_end >= 10, "Disparity is not available for this frame range"
+        if "flow_occ" in load or "flow_noc" in load:
+            assert sequence_start <= 10 and sequence_end >= 10, "Flow is not available for this frame range"
+
```
jsalotti, line 46: Ground-truth is computed for the gray camera. If ground-truth (disp or flow) is asked, make sure `grayscale==True`

alodataset/kitti_stereo_flow_sflow2015.py
```diff
@@ -0,0 +1,385 @@
...
+            "flow_noc",
+            "scene_flow",
+            "obj_map",
+        ],
+        **kwargs,
+    ):
+        super().__init__(name=name, **kwargs)
+        self.sequence_start = sequence_start
+        self.sequence_end = sequence_end
+        self.grayscale = grayscale
```
jsalotti, line 35: The 2015 dataset is created from RGB images. We do not need to provide grayscale in this dataset class.

alodataset/kitti_stereo_flow_sflow2015.py
```diff
@@ -0,0 +1,385 @@
...
+                sequence[index]["right"].baseline = ["baseline"]
+                sequence[index]["right"].append_cam_intrinsic(calib["right_intrinsic"])
+                sequence[index]["right"].append_cam_extrinsic(calib["right_extrinsic"])
+
+            # Frames at index 10 and 11 are the only one who have ground truth in dataset.
+            if index == 11:
+                if "disp_noc" in self.load:
+                    sequence[11]["left"].append_disparity(
+                        self.load_disp(os.path.join(self.split_folder, f"disp_noc_1/{idx:06d}_10.png")), "disp_noc"
+                    )
```
jsalotti, line 216: Reading the disparity for image 11 in this way will cause problems, because it is warped to be expressed in the reference frame of image 10. 
This goes against the current convention in alodataset, that the disparity attached to a `Frame` object is expressed in the same coordinate system.

jsalotti (changes requested):
Review for stereo 2012 and 2015.

alodataset/kitti_stereo_flow_sflow2015.py
Ardorax: The grayscale images are provided as an extension. So I just change the parameter as False by default and verify no ground truth is loaded when grayscale is True.

alodataset/kitti_stereo_flow_sflow2015.py
```diff
@@ -0,0 +1,385 @@
...
+                    sequence[10]["left"].append_flow(
+                        self.load_flow(os.path.join(self.split_folder, f"flow_occ/{idx:06d}_10.png")), "flow_occ"
+                    )
+                if "flow_noc" in self.load:
+                    sequence[10]["left"].append_flow(
+                        self.load_flow(os.path.join(self.split_folder, f"flow_noc/{idx:06d}_10.png")), "flow_noc"
+                    )
+                if "scene_flow" in self.load:
+                    # The scene flow cannot be added currently because of a bug in his representation.
+                    scene_flow = self.scene_flow_from_disp(sequence[10]["left"], sequence[11]["left"])
```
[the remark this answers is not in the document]
Ardorax: The bug is due to the fact that scene flow doesn't have the __get_view__ method.

alodataset/kitti_stereo_flow_sflow2015.py
jsalotti: @thibo73800 i'm not sure about what should be done here. 
- a user familiar with  KITTI Stereo2015 would expect the disparity of t1 to be warped to t0
- a user familar with aloception would expect `frame_t1.disp` to be the disparity of t1 (not warped)

We could append the disparity to the frame with an explicit name like `"warped_disparity"` to make sure it is not mistakenly used as a normal disparity map.

If we wanted to compute the unwarped disparity, we would need the flow for t1 to t0 (not given) or compute it with a pseudo-inversion (dirty) : none of these solutions seem satisfying.

alodataset/kitti_stereo_flow_sflow2015.py
```diff
@@ -0,0 +1,385 @@
...
+        data.update(
+            self._load_calib_cam_to_cam(
+                os.path.join(path, "calib_velo_to_cam", f"{idx:06d}.txt"),
+                os.path.join(path, "calib_cam_to_cam", f"{idx:06d}.txt"),
+            )
+        )
+
+        # Return the parameters we want only
+        result = {
+            "baseline": data["b_gray"] if self.grayscale else data["b_rgb"],
```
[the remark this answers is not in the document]
Ardorax: Grayscale is now an option not the default parameter.

alodataset/kitti_stereo_flow_sflow2015.py
thibo73800: I would suggest to start with warped_disparity for now

Ardorax: I named the disparity `warped_disp_noc` and `warped_disp_occ`.

jsalotti (changes requested):
Review kitti_object

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,288 @@
...
+            self.seq_params["baseline"] = calib["b_gray" if self.grayscale else "b_rgb"]
+            self.seq_params["left_intrinsic"] = CameraIntrinsic(
+                np.c_[calib[f"K_cam{0 if grayscale else 2}"], np.zeros(3)]
+            )
+            self.seq_params["left_extrinsic"] = CameraExtrinsic(calib[f"T_cam{0 if grayscale else 2}_rect"])
+            if right_frame:
+                self.seq_params["right_intrinsic"] = CameraIntrinsic(
+                    np.c_[calib[f"K_cam{1 if grayscale else 3}"], np.zeros(3)]
+                )
+                self.seq_params["right_extrinsic"] = CameraExtrinsic(calib[f"T_cam{1 if grayscale else 3}_rect"])
```
jsalotti, line 66: A different calibration file is read for each sequence. 
Therefore, `seq_params` should be indexed by the sequence name.

For example :
```
# loading the left intrinsic matrix for the sequence `seq`
cam_intrinsic = self.seq_params[seq]["left_intrinsic"]
```

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,288 @@
...
+                        )
+                    ).temporal()
+                )
+
+        frames = {}
+        frames["left"] = torch.cat(left, dim=0)
+        frames["left"].timestamp = item["times"]
+        frames["left"].baseline = self.seq_params["baseline"]
+        frames["left"].append_cam_intrinsic(self.seq_params["left_intrinsic"])
+        frames["left"].append_cam_extrinsic(self.seq_params["left_extrinsic"])
```
jsalotti, line 191: See the previous comment. The calibration (baseline, intrinsic, extrinsic) are different for different sequences. You should use the `sequence` variable to read these values.

jsalotti (changes requested):

alodataset/kitti_stereo_flow_sflow2015.py
```diff
@@ -0,0 +1,411 @@
...
+        # Compute depth from disparity at time T+1
+        disparity = next_frame.disparity["warped_disp_noc"]
+        next_depth = disparity.as_depth(baseline=0.54, camera_intrinsic=next_frame.cam_intrinsic)
+        next_depth.occlusion = disparity.mask.clone()
+        next_depth.append_cam_extrinsic(next_frame.cam_extrinsic)
+
+        # Remove the temporal dimension so depth and next_depth have the same shape.
+        next_depth = next_depth.squeeze(0)
+
+        scene_flow = SceneFlow.from_optical_flow(frame.flow["flow_noc"], depth, next_depth, frame.cam_intrinsic)
```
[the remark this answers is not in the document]
Ardorax: I forgot the disparity was warped.

thibo73800:
Is still a draft or a final pull request ?

jsalotti:
> Is still a draft or a final pull request ?

It is still a draft. It will be ready to merge by next week.

alodataset/kitti_tracking.py
```diff
@@ -20,6 +21,10 @@ def sequence_indices(n_samples, seq_size, skip, seq_skip):
         yield sequence_index(start, seq_size, skip)
 
 
+def default_value():
+    return []
+
+
 class KittiTrackingDataset(BaseDataset, SplitMixin):
```
jsalotti, line 28: no need to define this function

alodataset/kitti_tracking.py
```diff
@@ -68,19 +73,13 @@ def __init__(
                 self.seq_params[seq]["right_extrinsic"] = calib["right_extrinsic"]
 
             with open(os.path.join(self.dataset_dir, "label_02", f"{seq}.txt"), "r") as f:
-                labels = []
+                labels: defaultdict[int, List[List[str]]] = defaultdict(default_value)
```
jsalotti, line 76: `labels = defaultdict(list)`

alodataset/kitti_tracking.py
```diff
@@ -213,8 +195,9 @@ def getitem(self, idx: int) -> Dict[str, Frame]:
             bounding_box = BoundingBoxes2D(boxes2d, boxes_format="xyxy", absolute=True, frame_size=left_frame.HW)
             bounding_box.append_labels(labels, "track_id")
             bounding_box.append_labels(Labels(categories, labels_names=LABELS), "categories")
-            boxe3d = BoundingBoxes3D(boxes3d)
-            # TODO : boxes3D should also have labels
+            boxe3d = BoundingBoxes3D(
+                boxes3d, labels={"track_id": labels, "categories": Labels(categories, labels_names=LABELS)}
```
jsalotti, line 199: A same instance of `Labels` should not be the child of two different AugmentedTensor instance. Here `labels` is added as a child to both `bounding_box` and `boxe3d`. You should create a copy of it with `labels.clone()` instead of using the same instance.

Also, you have a condition for the bounding boxes 3D (you add the box only if the category is not "DontCare"). It is possible for some objects you only add a box 2D but not 3D. Therefore, the `Labels` objects for boxes2D and boxes3D are not exactly the same.

Is there a reason why you remove "DontCare" only for 3D boxes ?

jsalotti (changes requested):

alodataset/kitti_tracking.py
Ardorax: The objects with label `DontCare` doesn't have a correct box3d.

alodataset/kitti_road.py
```diff
@@ -0,0 +1,173 @@
...
+            for name, color_code in zip(names, color_codes):
+                area = ~np.array(ground_truth[:, :, color_code[0]] == color_code[1])
+                area = np.expand_dims(area, axis=0)
+                area = Mask(area, names=("C", "H", "W"))
+                frames["left"].append_mask(area, name=name)
+
+        return frames
+
+    # https://github.com/utiasSTARS/pykitti/tree/master
+    def read_calib_file(self, filepath):
```
Data-Iab, line 86: This function is repeated in all your classes with almost no changes, this is a good reason to create a parent class where you can add all repeated functions. Make other classes inherit from it to avoid functions redefinition. (with some small changes maybe)

Data-Iab (changes requested):

alodataset/kitti_object.py
```diff
@@ -0,0 +1,236 @@
+import os
+import numpy as np
+import torch
+from typing import Dict
+
+from alodataset import BaseDataset, Split, SplitMixin
+from aloscene import Frame, BoundingBoxes2D, Labels, BoundingBoxes3D
+from aloscene.camera_calib import CameraIntrinsic, CameraExtrinsic
+
+LABELS = ["Car", "Van", "Truck", "Pedestrian", "Person_sitting", "Cyclist", "Tram", "Misc", "DontCare"]
```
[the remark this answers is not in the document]
Ardorax: Is the new version correct ?

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,290 @@
...
+from alodataset import BaseDataset, SplitMixin, Split
+from aloscene import Frame, Pose, CameraIntrinsic, CameraExtrinsic
+
+
+def sequence_index(start, seq_size, skip):
+    end = start + (seq_size - 1) * (skip + 1)
+    return np.linspace(start, end, seq_size).astype(int).tolist()
+
+
+def sequence_indices(n_samples, seq_size, skip, seq_skip):
```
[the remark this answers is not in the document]
Ardorax: No, it's use by method `__init__`.

alodataset/kitti_depth.py
```diff
@@ -0,0 +1,240 @@
...
+    def __init__(
+        self,
+        subset: str = "all",
+        return_depth: bool = True,
+        custom_drives: Union[Dict[str, List[str]], None] = None,
+        name: str = "kitti",
+        main_folder: str = "image_02",
+        sample: bool = False,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
Ardorax: Black formater doesn't follow this rule. This change therefore can't be made.

alodataset/kitti_depth.py
```diff
@@ -0,0 +1,240 @@
...
+        Returns:
+            corresponding path to depth groud truth image.
+        """
+        fixed_dir = os.path.join("proj_depth", "groundtruth")
+        filename = os.path.basename(path)
+        driveAndSubset = os.path.join(*path.split(os.sep)[-5:-3])
+        return os.path.join(self.dataset_dir, "depth", driveAndSubset, fixed_dir, self.main_folder, filename)
+
+    def _get_paths_from_subdirs(
+        self, drive_path: str, filenames: Union[List[str], None] = None, extension: str = "png"
```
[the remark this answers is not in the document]
Ardorax: Black formater doesn't follow this rule. This change therefore can't be made.

alodataset/kitti_odometry.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(
+        self,
+        name="kitti_odometry",
+        sequences=None,
+        grayscale=True,
+        right_frame=True,
+        sequence_size=3,
+        skip=1,
+        sequence_skip=1,
+        **kwargs,
```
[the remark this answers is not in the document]
Ardorax: Black formater doesn't follow this rule. This change therefore can't be made.

alodataset/kitti_stereo_flow2012.py
```diff
@@ -0,0 +1,274 @@
...
+            "right",
+            "disp_noc",
+            "disp_occ",
+            "disp_refl_noc",
+            "disp_refl_occ",
+            "flow_occ",
+            "flow_noc",
+        ],
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
Ardorax: Black formater doesn't follow this rule. This change therefore can't be made.

alodataset/kitti_tracking.py
```diff
@@ -0,0 +1,324 @@
...
+    def __init__(
+        self,
+        name="kitti_tracking",
+        sequences: Union[int, List[int], None] = None,
+        right_frame=True,
+        sequence_size=3,
+        skip=1,
+        sequence_skip=1,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
Ardorax: Black formater doesn't follow this rule. This change therefore can't be made.

alodataset/kitti_semantic.py
```diff
@@ -0,0 +1,124 @@
...
+        super().__init__(name=name, **kwargs)
+
+        if self.sample:
+            raise NotImplementedError("Sample mode is not implemented for KittiSemanticDataset")
+
+        self.split_folder = os.path.join(self.dataset_dir, self.get_split_folder())
+        left_img_folder = os.path.join(self.split_folder, "image_2")
+        n_samples = len([file for file in os.listdir(left_img_folder)])
+
+        self.category = [
```
Data-Iab, line 25: Not practical. Can you store this in a json or yaml then load ?

Data-Iab (changes requested):

thibo73800 (approved):

jsalotti (approved):

PR #215, opened 2022-08-24
Optional trt import

thibo73800:
* ***Fix  optional trt import*** : Fix an import issue in the calibrator. All trt and prod package must be optional

_____
This pull request includes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)


Data-Iab (commented):
I think that we can isolate prod packages just by removing [this line](https://github.com/Visual-Behavior/aloception/blob/master/alonet/__init__.py#L14), can't we ?

PR #220, opened 2022-09-06
setup file

tflahaul:
Added a `setup.py` to facilitate the installation of aloception.

* ***setup.py***

To install, run one of:
```sh
pip install git+https://github.com/Visual-Behavior/aloception
```
```sh
python aloception/setup.py install
```
Then:
```python
>>> import aloscene
>>> 
```

_____
This pull request includes


setup.py
```diff
@@ -0,0 +1,36 @@
+from setuptools import setup, find_packages
+
+setup(
+	name='aloception',
+	author='Visual Behavior',
+	version='9.6.22',
```
Ardorax, line 6: Why this version number ?

tflahaul: Today's date, because there are no `version.txt` file yet

Ardorax: Maybe you should use the version of the current release of aloception.

setup.py
```diff
@@ -0,0 +1,34 @@
...
+        'more-itertools',
+        'tensorflow',
+        'pytorch-lightning',
+        'pycocotools',
+        'click',
+        'nvidia-tensorrt',
+        'onnx',
+        'onnxsim',
+        'pycuda',
+        'wandb'],
```
Ardorax, line 26: This package list is valid for aloception to work. However we get a lot of warning form torch because the version we currently use is an older one.

You should read the file `requirements.txt` in order to get all the needed package and their version.
Maybe you will encounter a problem with numpy. If it's the case you should add `numpy==1.22.2` to `requirements.txt` (this is the version used on the server). I made a similar patch for github action.

The last problem is about pytorch. This package should be installed by the user ([documentation](https://visual-behavior.github.io/aloception/getting_started/getting_started.html#install)) but he doesn't know which version.
**Maybe** it's possible to make the user choose the version he want (cuda, cpu...). If needed you can find all the version number of torch related package on the github action file: `.github/workflows/action.yml`.

Ardorax (changes requested):

thibo73800:
Thanks @tflahaul for the pull request. @Ardorax  Feedbacks are great. Once the changed are made according to @Ardorax suggestions we can merge this pull request.

setup.py
thibo73800: Along the way, we can also add the required version in the install section in the README.

Data-Iab (approved):

Issue #241, opened 2022-11-16
mergeable properties for augmented tensors


tflahaul:
As of today we can only merge tensors that have the exact same properties. Trying to merge tensors with different `.projection` or `.distortion` will raise
```sh
AssertionError: Trying to merge augmented tensor with different property: projection
```
<br/>

Being able to merge these properties in lists along with the childs would be great.

Expected return of a `torch.cat((a, b, c), dim=0)` with a, b & c being tensors with different projection properties:
```python
tensor(
	projection=['kumler_bauer', 'equidistant', 'pinhole']
	normalization=minmax_sym
	pose=torch.Size([3, 4, 4])
	mask=torch.Size([3, 1, H, W])
	depth=torch.Size([3, 1, H, W])
	cam_intrinsic=torch.Size([3, 4, 4])
	cam_extrinsic=torch.Size([3, 4, 4])
	distortion=[[0.45, 0.39, 0.12], 0.25, 1.0]
```

PR #244, opened 2022-11-17
alobugdays/241-aten-w-mergeable-properties

tflahaul:
Merge augmented tensors properties into lists instead of throwing assertion errors.
Fix issue #241 

`_merge_tensor`'s behavior is unchanged when the properties are the same.

### Example when using `torch.cat((a, b, c), dim=0)` with a, b, c being tensors with DIFFERENT projection properties
Before:
```sh
AssertionError: Trying to merge augmented tensor with different property: projection
```
After:
```python
tensor(
	projection=['kumler_bauer', 'equidistant', 'pinhole']
	normalization=minmax_sym
	pose=torch.Size([3, 4, 4])
	mask=torch.Size([3, 1, H, W])
	depth=torch.Size([3, 1, H, W])
	cam_intrinsic=torch.Size([3, 4, 4])
	cam_extrinsic=torch.Size([3, 4, 4])
	distortion=[[0.45, 0.39, 0.12], 0.25, 1.0]
        ...
```

### Example when using `torch.cat((a, b, c), dim=0)` with a, b, c being tensors with SAME projection properties
Before and after (the behavior is unchanged):
```python
tensor(
	projection=kumler_bauer
	normalization=minmax_sym
	pose=torch.Size([3, 4, 4])
	mask=torch.Size([3, 1, H, W])
	depth=torch.Size([3, 1, H, W])
	cam_intrinsic=torch.Size([3, 4, 4])
	cam_extrinsic=torch.Size([3, 4, 4])
	distortion=[0.45, 0.39, 0.12]
        ...
```
_____
This pull request includes

### DIFF
--- a/aloscene/bounding_boxes_2d.py
+++ b/aloscene/bounding_boxes_2d.py
@@ -33,7 +33,7 @@ class BoundingBoxes2D(aloscene.tensors.AugmentedTensor):
     x: list | torch.Tensor | np.array
         BoundingBoxes2D data. See explanation above for details.
     boxes_format: str
-        One of `xyxy` (y_min, x_min, y_max, x_max), `yxyx` (x_min, y_min, x_max, y_max) or `xcyc` (xc, yc, width,
+        One of `xyxy` (x_min, y_min, x_max, y_max), `yxyx` (y_min, x_min, y_max, x_max) or `xcyc` (xc, yc, width,
         height)
     absolute: bool
         Whether your boxes are encoded as absolute value or relative values (between 0 and 1). If absolute is True,
--- a/aloscene/tensors/augmented_tensor.py
+++ b/aloscene/tensors/augmented_tensor.py
@@ -450,14 +450,20 @@ class AugmentedTensor(torch.Tensor):
         # Setup the new structure before to merge
         prop_name_to_value = {}
 
+        visited = set()
         for tensor in tensor_list:
-
             for prop in tensor._property_list:
-                if prop in prop_name_to_value:
-                    assert prop_name_to_value[prop] == getattr(
-                        tensor, prop
-                    ), f"Trying to merge augmented tensor with different property: {prop}, {prop_name_to_value[prop]}, {getattr(tensor, prop)}"
-                prop_name_to_value[prop] = getattr(tensor, prop)
+                if prop in prop_name_to_value and prop_name_to_value[prop]:
+                    item = getattr(tensor, prop)
+                    if item != prop_name_to_value[prop]:  # different properties: merging is required
+                        if prop not in visited:
+                            prop_name_to_value[prop] = [prop_name_to_value[prop], item]
+                            visited.add(prop)
+                        else:
+                            prop_name_to_value[prop].append(item)
+                        setattr(n_tensor, prop, prop_name_to_value[prop])  # update
+                else:
+                    prop_name_to_value[prop] = getattr(tensor, prop)
 
             for label_name in tensor._children_list:
                 label_value = getattr(tensor, label_name)


thibo73800:
This is an important PR
@tflahaul  can you give more context about the expected behavior with one or two example ?
@jsalotti  you also had multiple time the same issue

thibo73800 (changes requested):
This is an important PR
@tflahaul can you give more context about the expected behavior with one or two example ?
@jsalotti you also had multiple time the same issue

tflahaul:
@thibo73800 I gave more explanations & a minimal example in issue #241

thibo73800:
@jsalotti  @tflahaul  @anhtu293 

The following work because there is indeed no conflicts between the properties
```
  f1 = aloscene.Frame(np.zeros((3, 200, 300)))
  f2 = aloscene.Frame(np.zeros((3, 200, 300)))
  f1.baseline = 0.5
  f2.baseline = 0.5
  c = torch.cat([f1.batch(), f2.batch()], dim=0)
```

The following will not work. But this is an expected behavior. Otherwise we could insert hidden issue in the pipeline.
```
f1.baseline = 0.5
f2.baseline = 0.6
c = torch.cat([f1.batch(), f2.batch()], dim=0)
```

An alternative way to do could be to create a new augmented tensor (Not a default child but a one that could be create into a specific datapipeline)
```
 baselines_1 = aloscene.tensors.AugmentedTensor(torch.as_tensor([0.5]), names=("N",))
 baselines_2 = aloscene.tensors.AugmentedTensor(torch.as_tensor([0.6]), names=("N",))
 f1.add_child("baselines", baselines_1, align_dim=["B","T"], mergeable=True)
 f2.add_child("baselines", baselines_2, align_dim=["B","T"], mergeable=True)
 c = torch.cat([f1.batch(), f2.batch()], dim=0)
```

I'm don't like the ideas of simply doing a concat on the properties if they're different since it could break some other mechanism related to child and augmented tensors (slice for example will not work for properties). An alternative could be to automatically create a new child as above when trying to merge tensors with different properties. But I don't think it must be the default behavior.

jsalotti:
> @jsalotti @tflahaul @anhtu293
> 
> The following work because there is indeed no conflicts between the properties
> 
> ```
>   f1 = aloscene.Frame(np.zeros((3, 200, 300)))
>   f2 = aloscene.Frame(np.zeros((3, 200, 300)))
>   f1.baseline = 0.5
>   f2.baseline = 0.5
>   c = torch.cat([f1.batch(), f2.batch()], dim=0)
> ```
> 
> The following will not work. But this is an expected behavior. Otherwise we could insert hidden issue in the pipeline.
> 
> ```
> f1.baseline = 0.5
> f2.baseline = 0.6
> c = torch.cat([f1.batch(), f2.batch()], dim=0)
> ```
> 
> An alternative way to do could be to create a new augmented tensor (Not a default child but a one that could be create into a specific datapipeline)
> 
> ```
>  baselines_1 = aloscene.tensors.AugmentedTensor(torch.as_tensor([0.5]), names=("N",))
>  baselines_2 = aloscene.tensors.AugmentedTensor(torch.as_tensor([0.6]), names=("N",))
>  f1.add_child("baselines", baselines_1, align_dim=["B","T"], mergeable=True)
>  f2.add_child("baselines", baselines_2, align_dim=["B","T"], mergeable=True)
>  c = torch.cat([f1.batch(), f2.batch()], dim=0)
> ```
> 
> I'm don't like the ideas of simply doing a concat on the properties if they're different since it could break some other mechanism related to child and augmented tensors (slice for example will not work for properties). An alternative could be to automatically create a new child as above when trying to merge tensors with different properties. But I don't think it must be the default behavior.

I agree. I specifically don't like the idea of silently merging a property that should not be merged.

tflahaul:
Can the projection and distortion properties of atens already be expressed as childs ? @jsalotti @thibo73800 
Or is there some work to do to make it work

jsalotti:
> Can the projection and distortion properties of atens already be expressed as childs ? @jsalotti @thibo73800 Or is there some work to do to make it work

Distortion should be OK, because it is numerical. Projection is more complicated because they are string.
This could be done by replacing it by int :

```
Class Projection :
    PINHOLE = 0
    EQUIDISTANT = 1
    KUMLER_BAUER= 2
    
```
and testing against it

thibo73800:
@jsalotti  Agree, also if one want to link the text with the child we have the `aloscene.Labels` property that can store the name associated with each ID. Labels could be used to store projection.

jsalotti:
I still think that it is a dangerous idea to merge with different properties.

For example with projection, the function `depth.as_point3d` has different behavior based on `projection` property. If the user merges two tensors with different values for projection, the method `as_point3d` will not work for the merged tensor.

But maybe in some cases it still is absolutely necessary to merge such tensor, even if it break some other behavior ? @tflahaul could you cite one scenario where it is the case ?

jsalotti:
@tflahaul the idea behind your code sample seems good, but I think we will encounter some problems. These problems are caused by the difference between `property` and `child`. 

Properties lack some interesting features that children have : a child is an `AugmentedTensor` with names for its dimensions. It can be mergeable or not, and the `align_dim` argument gives us flexibility on how the child should be merged. This allows others functions to operate seemlessly on `AugmentedTensor` whether they have batch/temporal dimension or not.

Some example following the idea of your code sample :
```py
f1 = Frame(torch.ones((1,3,10,10)), names=("T","C","H", "W"))
f2 = f1.clone()
f1.projection = "pinhole"
f2.projection = "kumler_bauer"
f = torch.cat([f1,f2], axis=0) # concat along temporal axis
f = f.batch() # add batch dimension
print(f.projection)
print(f.names)
```
If we follow the idea of creating a list for properties when merging, the output will be:
```py
['pinhole', 'kumler_bauer']
('B', 'T', 'C', 'H', 'W')
```
Then, the projection function would have to guess that the two items of the list correspond to the temporal dimension, which correspond to `dim=1`.

I think the behavior you want to having is much closer to what `child`, as an `AugmentedTensor` can do. I would suggest to encode the projection as and `AugmentedTensor` of `int`, and define it as a child instead of a property.

thibo73800:
@jsalotti  @tflahaul  I think the best is to used Labels
```python
aloscene.Labels([0, 1, 1, 2], labels_names=["stereo", "kumer", "equidistant"], names=("N",))
```

Data-Iab:
closing since no activity since November

Issue #38, opened 2021-09-07
alodataset Sintel datasets : add test split


jsalotti:
For the moment, only data from the train split can be loaded in: 
- `SintelFlowDataset`
- `SintelDisparityDataset`
- `SintelMultiDataset`

The test split should be added. An error or at least warning should be raised if the user tries to load the dataset with labels (unavailable in test split).

thibo73800:
Is it quick to do ? If not remove the quick-fix label please

jsalotti:
I think it should be quick. But test data is only images, as ground-truth is kept secret by the authors.

PR #265, opened 2022-11-17
Alobugdays : added "training" or "testing" param

Dee61298:
Solves issue #38 
Minimal reproducible example is in sintel_disparity_dataset.py in main() 

However, I wasn't able to finish the tests because the testing split is missing from /data. @ragier I heard you were the one responsible for the datas, so, should I download it again ?

### DIFF
--- a/alodataset/sintel_base_dataset.py
+++ b/alodataset/sintel_base_dataset.py
@@ -2,13 +2,13 @@ import itertools
 import torch
 import os
 
-from aloscene.io.disparity import load_disp_png
+from alodataset import BaseDataset, SequenceMixin, Split
 from aloscene import Frame, Flow, Mask, Disparity
+from aloscene.io.disparity import load_disp_png
 from aloscene.utils.data_utils import DLtoLD
-from alodataset import BaseDataset, SequenceMixin
 
 
-class SintelBaseDataset(BaseDataset, SequenceMixin):
+class SintelBaseDataset(BaseDataset, Split, SequenceMixin):
     """
     Abstract Base Class for MPI Sintel datasets
 
@@ -55,7 +55,7 @@ class SintelBaseDataset(BaseDataset, SequenceMixin):
         "temple_3",
     ]
 
-    def __init__(self, training=True,cameras=None, labels=None, passes=None, sintel_sequences=None, name="Sintel_base", **kwargs):
+    def __init__(self, cameras=None, labels=None, passes=None, sintel_sequences=None, name="Sintel_base", **kwargs):
         super(SintelBaseDataset, self).__init__(name=name, **kwargs)
 
         if self.sequence_skip != 0:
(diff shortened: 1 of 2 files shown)


alodataset/sintel_base_dataset.py
```diff
@@ -55,7 +55,7 @@ class SintelBaseDataset(BaseDataset, SequenceMixin):
         "temple_3",
     ]
 
-    def __init__(self, cameras=None, labels=None, passes=None, sintel_sequences=None, name="Sintel_base", **kwargs):
+    def __init__(self, training=True,cameras=None, labels=None, passes=None, sintel_sequences=None, name="Sintel_base", **kwargs):
```
anhtu293, line 58: we support Split dataset in `alodataset`. It's cleaner to  use that for train/val split.

```
from alodataset import BaseDataset, Split, SplitMixin

class SintelBaseDataset(BaseDataset, Split, SequenceMixin):

SPLIT_FOLDERS = {Split.VAL: "validation", Split.TRAIN: "training"}
path = os.path.join(dset_dir, self.split, ...)
```

In that code `self.split` is `Split.VAL` or `Split.TRAIN`

alodataset/sintel_base_dataset.py
```diff
@@ -67,8 +67,11 @@ def __init__(self, cameras=None, labels=None, passes=None, sintel_sequences=None
         self.sintel_sequences = sintel_sequences if sintel_sequences is not None else self.SINTEL_SEQUENCES
         if self.sample:
             return
-
         self._assert_inputs()
+        if training:
+            self.split = "training"
+        else:
+            self.split = "testing"
```
anhtu293, line 74: with `Split` class, we dont need `training` argument anymore

alodataset/sintel_base_dataset.py
```diff
@@ -98,7 +101,7 @@ def _split_cameras(self, my_dict):
 
     def _get_folder(self, sintel_seq, feature_or_label):
         dset_dir = self.dataset_dir
-        return os.path.join(dset_dir, "training", feature_or_label, sintel_seq)
+        return os.path.join(dset_dir, self.split, feature_or_label, sintel_seq)
```
anhtu293, line 104: same

alodataset/sintel_disparity_dataset.py
```diff
@@ -105,7 +105,10 @@ def _get_camera_frames(self, sequence_data, camera):
 
 
 if __name__ == "__main__":
-    dataset = SintelDisparityDataset(sample=True)
+    dataset = SintelDisparityDataset(sample=False, training=False)
```
anhtu293, line 108: it's better to use `alodataset.Split`

alodataset/sintel_base_dataset.py
Dee61298: Why can't it be `Split.TEST` since that's the point of issue #38 ?

anhtu293: it can be `Split.TEST` for test dataset

Data-Iab (approved):

Issue #121, opened 2021-10-20
BoundingBoxes does not show the appropriate colors


Johansmm:
For the code:

```python
from alodataset import CocoDetectionDataset

coco_dataset = CocoDetectionDataset(img_folder="val2017", ann_file="annotations/instances_val2017.json")
for f, frames in enumerate(coco_dataset.train_loader(batch_size=2)):
    frames = Frame.batch_list(frames)
    frames.get_view().render()
    if f > 1:
        break
```

Each image rendered show all boxes with the same color, but the ID is shown with the correct color

PR #277, opened 2022-11-17
Alobugdays : added warning message for labels_set

Dee61298:
Solves issue #121 
The fix is actually quite simple, you just need to call 
`frames.get_view(labels_set=NAME_OF_YOUR_LABELS_SET).render()`

instead of 
`frames.get_view().render()
`
I added an explanatory messages to BoundingBoxes2d.get_view() with the available labels so the user finds the solution more easily


Data-Iab:
Solved [here](https://github.com/Visual-Behavior/aloception-oss/pull/282)

PR #278, opened 2022-11-17
Alobugdays : changed order of date in run_id

Dee61298:
Changed expe_name to have year first & month as a number instead of string


Data-Iab:
The order correct as in American English, changing it will lead to folders with various naming conventions.

Issue #21, opened 2021-09-07
Convert frame to an uint8 numpy array


thibo73800:
The following code should be available as a method of Frame instances, in order to convert the frame to numpy array in uint8 image format.

**Code example**

```
frame = frame.detach()
frame = frame.norm255()
frame = frame.cpu()
frame = frame.type(torch.uint8)
frame = frame.rename(None)
frame = frame.permute([1, 2, 0])
frame = frame.contiguous()
frame = frame.numpy()
```

thibo73800:
This can be implement on top of `as_numpy()`

PR #280, opened 2022-11-17
Alobugdays : Add method to convert frame to uint8 numpy array

eleonorefrcs:
Issue: #21 

***Convert frame to uint8 numpy array*** : Add method in `Frame` class to convert tensor in numpy array.

_____
This pull request includes

### DIFF
--- a/aloscene/frame.py
+++ b/aloscene/frame.py
@@ -350,6 +350,22 @@ class Frame(aloscene.tensors.SpatialAugmentedTensor):
         """
         self._append_child("scene_flow", scene_flow, name)
 
+    def as_numpy(self, dtype=torch.uint8):
+        """Convert the frame to numpy array uint8 format.
+
+        Parameters
+        ----------
+            dtype: dtype or str
+                The output dtype. Default torch.uint8.
+
+        Returns
+        -------
+            np.ndarray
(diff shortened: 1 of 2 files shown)


Data-Iab:
we have it already [here](https://github.com/Visual-Behavior/aloception-oss/blob/master/aloscene/tensors/augmented_tensor.py#L695)

Data-Iab:
Thanks @eleonorefrcs for reopening.
As the method `as_numpy` is already in AugTensors and used in many lines of our code, overriding it is not a good idea. I suggest to rename the function to `as_image`.

aloscene/frame.py
```diff
@@ -350,6 +350,22 @@ def append_scene_flow(self, scene_flow: SceneFlow, name: Union[str, None] = None
         """
         self._append_child("scene_flow", scene_flow, name)
 
+    def as_numpy(self, dtype=torch.uint8):
```
Data-Iab, line 353: Thanks @eleonorefrcs for reopening.
As the method `as_numpy` is already in AugTensors and used in many lines of our code, overriding it is not a good idea. I suggest to rename the function to `as_image`.

Data-Iab (changes requested):

PR #284, opened 2022-11-18
__getitem__ with augmentedTensor

ragier:
1. fix __getitem__ on augmentedTensor with augmented tensor as mask.
2. reset name only if tensor aren't linearized (like in bbox unit test) else declass to classic tensor


ragier:
#bug_compte_double

ragier:
1. Pass rotate image test
2. Pass bbox test

Data-Iab (approved):
seems ok, my tests passed as well

Issue #225, opened 2022-09-20
Box3d don't render correctly when a part of the box is outside the image


Ardorax:
Under certain condition, a BoundingBoxes3D can be incorrectly draw if a part of it is outside the image.
```py
import numpy as np
from aloscene import Frame, BoundingBoxes3D, CameraExtrinsic, CameraIntrinsic

frame = Frame(np.zeros((3, 275, 1242)))


intrisic = CameraIntrinsic(np.array([[721.5377, 0.0, 609.5593, 0.0], [0.0, 721.5377, 172.854, 0.0], [0, 0, 1, 0]]))
extrinsic = CameraExtrinsic(np.array([[1, 0, 0, 0.062169], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]))
frame.append_cam_intrinsic(intrisic)
frame.append_cam_extrinsic(extrinsic)
box3d = BoundingBoxes3D([[3.577114, 0.6670290000000001, 1.436819, 1.895275, 2.195312, 5.530314, 3.1704533267948967]])
frame.append_boxes3d(box3d)

# Box3D don't render correctly
frame.get_view().render()
```

Aurelien-VB:
Le bug persiste même avec tous les points dans l'image

![image](https://user-images.githubusercontent.com/114914605/202399926-5d87ad8f-29d9-44f0-a5be-ca4548ddbc8c.png)

```py
import numpy as np
from aloscene import Frame, BoundingBoxes3D, CameraExtrinsic, CameraIntrinsic

frame = Frame(np.zeros((3, 1000, 1242)))


intrisic = CameraIntrinsic(np.array([[721.5377, 0.0, 309.5593, 0.0], [0.0, 721.5377, 300.854, 0.0], [0, 0, 1, 0]]))
extrinsic = CameraExtrinsic(np.array([[1, 0, 0, 0.062169], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]))
frame.append_cam_intrinsic(intrisic)
frame.append_cam_extrinsic(extrinsic)
box3d = BoundingBoxes3D([[3.577114, 0.6670290000000001, 1.436819, 0.895275, 2.195312, 5.530314, 3.1704533267948967]])
frame.append_boxes3d(box3d)

# Box3D don't render correctly
frame.get_view().render()```

PR #291, opened 2022-11-18
Fix boxes3d for points with negative depth

Aurelien-VB:
This fixes issue #225, encountered when some points of boxes3d have negative depth (part of the object is behind the camera).

Points with negative depth are re-projected to positive depth by negating them. The line is then extend to outside the image to keep it coherent.

To extend the line, I compute the normalized vector from the linked point inside the frame to the point with negated depth (in image space). I then rescale the vector and add it to the negated point to extend outside the visible frame.

To find the right linked point, I take the furthest away from the point of negative depth. There is only 2 possible points since the third also has negative depth and thus there is no visible line that can can be seen from it. 

The only limitation of this technique is if a point has negative depth and linked to two points in the visible image then one of the line will be distorted.

I also fix NaN when depth = 0.

Code to replicate results:

```py
import numpy as np
from aloscene import Frame, BoundingBoxes3D, CameraExtrinsic, CameraIntrinsic

frame = Frame(np.zeros((3, 1000, 1000)))
intrisic = CameraIntrinsic(np.array([[500, 0.0, 500, 0.0], [0.0, 500, 500, 0.0], [0, 0, 1, 0]]))
extrinsic = CameraExtrinsic(np.array([[1, 0, 0, 0.062169], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]))
frame.append_cam_intrinsic(intrisic)
frame.append_cam_extrinsic(extrinsic)
box3d = BoundingBoxes3D([[-1, -1, 3, 1, 1, 10, 0], [1, 1, 3, 10, 1, 1, np.pi/3]])
frame.append_boxes3d(box3d)

frame.get_view([frame.boxes3d]).render()

```

### DIFF
--- a/aloscene/bounding_boxes_3d.py
+++ b/aloscene/bounding_boxes_3d.py
@@ -163,11 +163,47 @@ class BoundingBoxes3D(aloscene.tensors.AugmentedTensor):
         vertices = torch.cat([vertices, ones], dim=1)
         # vertices_2d = cam_intrinsic @ cam_extrinsic @ vertices # (n, 4, 8)
         vertices = cam_extrinsic @ vertices  # (n, 4, 8)
+
         if not include_vertice_behind_image_plan and vertices.shape[0] > 0:
             behind_plan_filter = vertices[:, 2, :].max(-1).values > 0
             vertices = vertices[behind_plan_filter]
+
+        # find values where depth is negative
+        neg_mask = vertices[:, 2] < 0.0
+
+        if torch.any(neg_mask):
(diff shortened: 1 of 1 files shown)


Aurelien-VB:
#225

Data-Iab:
After the quick discussion, can you make this PR as draft till you solve all its issues ? @Aurelien-VB

Data-Iab:
feel free to re-open when you have time for it :))

Issue #17, opened 2021-09-07
Recursive cropping of labels can lead to silent unwanted behaviors


thibo73800:
**Example**
Let us create a Frame, and attach to it a flow at half-resolution:

```
frame = Frame(np.random.uniform(0, 1, (3, 14, 14)), names=("C", "H", "W"))
flow  = Flow(np.random.uniform(0, 1, (2, 7, 7)), names=("C", "H", "W"))
frame.append_flow(flow)
```

Now, let us crop the frame as follow :
```
top, left = (1, 4)
h, w = (3, 4)
frame_cropped = frame[:, top:top+h, left:left+w]
```

Because the coordinates and the dimension of the cropped region are not multiple of 2, it is not possible to crop a region in the Flow object that correspond exactly. 

The exactly corresponding region in the flow coordinate system needs to be expressed with float coordinates:
```
flow_top, flow_left = (0.5, 2.)
flow_h, flow_w = (1.5, 2)
```

But in the current code, these float values are silently rounded to the nearest int, causing multiple problems :
- misalignement between frame coordinates and flow coordinates
- mismatch of aspect ratio between frame and flow objects

In this example :
```
print(f"frame_cropped:\t{frame_cropped.names} = {tuple(frame_cropped.shape)}")
print(f"frame_cropped.flow:\t{frame_cropped.flow.names} = {tuple(frame_cropped.flow.shape)}")
```
gives us the following output:
```
frame_cropped:	('C', 'H', 'W') = (3, 3, 4)
frame_cropped.flow:	('C', 'H', 'W') = (2, 2, 2)
```
where the cropped flow now has a square aspect-ratio, and its top-left corner does not correspond to the top-left corner of the cropped frame.

Multiple thoughts about it :
- this should not be allowed by default because it is misleading for users (and silent)
- we should at least print a pretty visible warning
- the user should be able to set a flag (on the frame or a global flag) to allow or forbid this behavior.

Some consequences :
- if the current behavior is not allowed, but frame and labels have different resolution, some operation won't work. In this example where the frame is 2x larger than flow, a random crop should know that the cropped region coordinates and dimension should be multiples of two. If it does not take this constraint into account, it will generate an error when trying to crop with uneven coordinates or dims.

ragier:
@thibo73800 What's your current mindset about this?

PR #294, opened 2022-11-18
17 Not integer crop within recursion

ragier:
#17 

To reproduce : 
```
from aloscene import Frame, Flow
import numpy as np

import warnings

frame = Frame(np.random.uniform(0, 1, (3, 14, 14)), names=("C", "H", "W"))
flow  = Flow(np.random.uniform(0, 1, (2, 7, 7)), names=("C", "H", "W"))
frame.append_flow(flow)

top, left = (1, 4)
h, w = (3, 4)


print(f"frame_cropped:\t{frame_cropped.names} = {tuple(frame_cropped.shape)}")
print(f"frame_cropped.flow:\t{frame_cropped.flow.names} = {tuple(frame_cropped.flow.shape)}")
```

__New behavior : Display warning when non-integer crop append__

To remove this specific warning : 
```
with warnings.catch_warnings():
    warnings.simplefilter('ignore', UserWarning)
    frame_cropped = frame[:, top:top+h, left:left+w]

```

### DIFF
--- a/aloscene/tensors/spatial_augmented_tensor.py
+++ b/aloscene/tensors/spatial_augmented_tensor.py
@@ -12,6 +12,7 @@ import aloscene
 from aloscene.camera_calib import CameraExtrinsic, CameraIntrinsic
 from aloscene.utils.data_utils import LDtoDL
 
+import warnings
 
 class SpatialAugmentedTensor(AugmentedTensor):
     """Spatial Augmented Tensor. Used to represets any 2D data. The spatial augmented tensor can be used as a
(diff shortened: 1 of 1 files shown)


Data-Iab:
@ragier Don't we need this warning in case of resizing to ?

ragier:
Resize isn't always non-integer?

The only case I can imagine is a half-resolution resize with nearest neighbor, but a good practice should be to use a strided_slice instead. (which isn"t implemented on frame, but doesn't make sense I think)

Data-Iab (approved):

PR #296, opened 2022-11-18
alobugdays : added asking permission to crowdhuman

Dee61298:
asking user for permission before preparing dataset


Data-Iab:
As long as the function `prepare` is not called inside any method we don't need to include any permissions. If the user calls it, it's because he wants to.

PR #304, opened 2022-11-18
allow batch_list to batch tensors with different children

### LINKED ISSUE #272: Choice of required properties during call to batch_list
# Description

When using a dataset from `alodataset`, the items of the dataset are instances of `aloscene.Frame` and have often a lot properties. When solving a problem, we might :

- want to use data from different datasets, that do not have the same properties
- be interested in only a subset of the properties, that is available for both dataset

## Simple example : merging two datasets

We want to train a flow model using data from multiple datasets.

### Original datasets

dataset1 : stereo dataset with flow
- flow and disparity properties are not `None`
- baseline property is not `None`

dataset2 : monocular flow
- only flow property ; disparity is `None`
- baseline is `None`

### Merged dataset
This can be done by creating an `alodataset.MergeDataset` to sample frames from both dataset. But when batching the frame, with `aloscene.batch_list`, we will encounter an error because of property `baseline` and child `disparity`.

```py
from torch.utils.data.sampler import Sampler
from alodataset import FlyingThings3DSubsetDataset, MergeDataset
import aloscene

d1 = FlyingThings3DSubsetDataset(labels=["flow", "disp", "flow_occ"])
d2 = FlyingThings3DSubsetDataset(labels=["flow"])
d = MergeDataset([d1, d2])
for data in d.train_loader(batch_size=10):
    break
data = aloscene.batch_list(data)
```

jsalotti:
Add an argument to `aloscene.batch_list`, allowing to batch tensors with different children or properties with different values. 
Closes #272 

* ***New feature 1*** : Merge with different properties
```python
f1 = Frame(torch.ones(3,10,10))
f2 = Frame(torch.ones(3,10,10))
f1.baseline = 0.5
f2.baseline = 1
f = aloscene.batch_list([f1, f2], intersection=True)
print(f.baseline)
```
Here, the `baseline` property has different values for different tensors. Therefore, the property baseline of the batched tensors has `baseline` set to `None`.

* ***New feature 2*** : Merge with different mergeable children
```python
f1 = Frame(torch.ones(3,10,10))
f2 = Frame(torch.ones(3,10,10))
f1.append_disparity(Disparity(torch.ones(1,10,10)))
f = aloscene.batch_list([f1, f2], intersection=True)
print(f.disparity)
```
Only one of the two frames has a disparity child, therefore the batched tensor will have `disparity` set to `None`

*Note: this will not drop unmergeable children, because by definition they allow that some tensors have `None` values for a child before merging* 
_____
This pull request includes

### DIFF
--- a/aloscene/__init__.py
+++ b/aloscene/__init__.py
@@ -21,8 +21,8 @@ from .renderer import Renderer
 from typing import Union
 
 
-def batch_list(tensors):
-    return SpatialAugmentedTensor.batch_list(tensors)
+def batch_list(tensors, intersection=False):
+    return SpatialAugmentedTensor.batch_list(tensors, intersection=intersection)
 
 
 _renderer = None
--- a/aloscene/tensors/augmented_tensor.py
+++ b/aloscene/tensors/augmented_tensor.py
@@ -544,7 +544,7 @@ class AugmentedTensor(torch.Tensor):
                     # if we removed all keys, set this child to None
                     if intersection and not labels_dict2list[label_name]:
                         labels_dict2list[label_name] = None
-                elif intersection and len(labels_dict2list[label_name]) != dim_size or (None in labels_dict2list[label_name]):
+                elif intersection and (len(labels_dict2list[label_name]) != dim_size or (None in labels_dict2list[label_name])):
                     labels_dict2list[label_name] = None
                 else:
                     args = list(args)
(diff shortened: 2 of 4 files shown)


thibo73800 (changes requested):
Can you provide this PR with a new set of unit test to check compatibility with previous behavior + exposing the different new possible scenarios ?

Data-Iab (approved):
Tested.
Will be merged once @thibo73800 approves

jsalotti:
@Data-Iab Unittests have been added ; and an error fixed. The branch is now ready to be merged.

Data-Iab (approved):

PR #308, opened 2022-11-28
kumler-bauer now support distortion coefficients in pixel instead of in meter

anhtu293:
General description of your pull request with the list of new features and/or bugs.

* ***New feature 1*** : Kumler-bauer distortion coefficients are now in pixel instead of meter. This is more logical with aloception computation because we work alot on image plan, it's too complicated to have a coefficient in meter. Since now, `distortion ` property in case of `kumler_bauer` only needs 2 elements instead of 3 in actual version.
```python
import numpy as np
from aloscene import Frame

data = np.random.uniform(size=(3, 1024, 1024))
frame = Frame(data, names=("C", "H", "W"), projection="kumler_bauer", distortion=[0.1, 0.1])

```
_____
This pull request includes

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/aloscene/depth.py
+++ b/aloscene/depth.py
@@ -292,7 +292,7 @@ class Depth(aloscene.tensors.SpatialAugmentedTensor):
                 focal_length = focal_length * theta * dist_coef / r.abs()
             elif projection == "kumler_bauer":
                 focal_length = (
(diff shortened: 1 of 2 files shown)


Data-Iab:
closing this ... :))

PR #310, opened 2022-12-08
Added title to get_view()

Dee61298:
- ***New feature 1*** : Added title to frames displayed with get_view()

Added a "title" argument to the get_view() method to be able to directly input a title for your display .  
  `        frames.get_view(title="test").render()
  `

- ***New feature 2*** : Added a default title to Matplotlib viewer
If you use the Matplotlib viewer (default) and don't pass a title, "Frame" is the default title for the display (copy of OpenCV viewer behavior)
`        frames.get_view().render()
`

This pull request includes

- [ ]  Bug fix (non-breaking change which fixes an issue)
- [X]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  This change requires a documentation update

### DIFF
--- a/aloscene/renderer/renderer.py
+++ b/aloscene/renderer/renderer.py
@@ -4,8 +4,6 @@ import math
 import matplotlib.pyplot as plt
 import matplotlib
 from abc import ABC, abstractmethod
-import numpy as np
-import cv2
 
 
 def adapt_text_size_to_frame(size, frame_size):
@@ -93,14 +91,16 @@ class View(object):
         if location is not None:
             plt.figure(figsize=figsize, tight_layout=True)
             plt.imshow(self.image)
-            plt.title("Frame" if self.title is None else self.title, self.image[:, :, ::-1])
+            if self.title is not None:
+                plt.title(self.title)
             plt.savefig(location)
             plt.close()
 
         if method == self.MATPLOTLIB:
             plt.figure(figsize=figsize, tight_layout=True)
             plt.imshow(self.image)
-            plt.title("Frame" if self.title is None else self.title, self.image[:, :, ::-1])
+            if self.title is not None:
+                plt.title(self.title)
             plt.show()
             plt.close()
 
--- a/aloscene/renderer/renderer.py
+++ b/aloscene/renderer/renderer.py
@@ -93,12 +93,14 @@ class View(object):
         if location is not None:
             plt.figure(figsize=figsize, tight_layout=True)
             plt.imshow(self.image)
+            plt.title("Frame" if self.title is None else self.title, self.image[:, :, ::-1])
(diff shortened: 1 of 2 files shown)


aloscene/renderer/renderer.py
```diff
@@ -93,12 +93,14 @@ def render(self, method="matplotlib", location: str = None, figsize=[6.4, 4.8]):
         if location is not None:
             plt.figure(figsize=figsize, tight_layout=True)
             plt.imshow(self.image)
+            plt.title("Frame" if self.title is None else self.title, self.image[:, :, ::-1])
```
Data-Iab, line 96: The second arg is a dict, I don't get why it's an array here ?

aloscene/renderer/renderer.py
```diff
@@ -93,12 +93,14 @@ def render(self, method="matplotlib", location: str = None, figsize=[6.4, 4.8]):
...
             plt.savefig(location)
             plt.close()
 
         if method == self.MATPLOTLIB:
             plt.figure(figsize=figsize, tight_layout=True)
             plt.imshow(self.image)
+            plt.title("Frame" if self.title is None else self.title, self.image[:, :, ::-1])
```
Data-Iab, line 103: The second arg is a dict, I don't get why it's an array here ?

aloscene/renderer/renderer.py
Data-Iab, line 96: "Frame" can not be a view title, I prefer to call `plt.title` only if self.title is not None to keep the previous behavior.

aloscene/tensors/spatial_augmented_tensor.py
```diff
@@ -130,7 +130,7 @@ def get_view(self, views: list = [], exclude=[], size=None, grid_size=None, **kw
         """
         _views = [v for v in views if isinstance(v, View)]
         if len(_views) > 0:
-            return View(Renderer.get_grid_view(_views, grid_size=None, cell_grid_size=size, **kwargs))
+            return View(Renderer.get_grid_view(_views, grid_size=None, cell_grid_size=size, **kwargs),title=title)
```
Data-Iab, line 133: keep space after `,`

Data-Iab (changes requested):

aloscene/renderer/renderer.py
Dee61298: Why not ? It is with the OpenCV viewer (https://github.com/Visual-Behavior/aloception-oss/blob/master/aloscene/renderer/renderer.py#L106) so I was trying to have the same behavior for both viewers

Data-Iab (approved):

PR #320, opened 2023-02-02
Some features that I needed when working on deformableDETR

Dee61298:
Some features that I needed when working on deformableDETR

* ***New feature 1*** : Added "sequential" arg to Data2Detr datamodule
Uses sequential sampler for train_dataloader instead of random sampler. 
If you run this command, you should first see some lunchboxes with some broccoli inside, and then a flower bouquet.

```
python data_modules/coco_detection2detr.py --sequential
```
* ***New feature 2*** : The layers are instanciated in order in DETR and DeformableDETR
Before, when running "print(model)" on either of the models, the layers would be displayed out of order, which was confusing (i.e transformer before backbone). Now they are displayed in order. 

```
from argparse import ArgumentParser
from alonet.detr import LitDetr

import alonet


def get_arg_parser():
    parser = ArgumentParser(conflict_handler="resolve")
    parser = alonet.common.add_argparse_args(parser)  # Common alonet parser
    parser = LitDetr.add_argparse_args(parser)  # LitDetr training parser
    # parser = pl.Trainer.add_argparse_args(parser) # Pytorch lightning Parser
    return parser


def main():
    """Main"""
    # Build parser
    args = get_arg_parser().parse_args()  # Parse

    # Init the Detr model with the dataset
    detr = LitDetr(args)
    print(detr.model)

if __name__ == "__main__":
    main()
```
* ***New feature 3*** : Added spatial_shift to mask
The function is empty for now because its early implementation would trigger unforeseen bugs in DeformableDETR. However, having the function is necessary, because it triggers an error message instead when using Spatial Shift for data augmentation. 
If you run this example without the modification, you'll get a "raise Exception(f"This Augmented tensor {type(self)} should implement this method")". If you run the example with the modification, you'll get no errors. 

```
import aloscene
import torch
import alodataset.transforms as T

x=torch.zeros(3,256,256)
frame=aloscene.Frame(x,names=("C","H","W"))
mask=aloscene.Mask(torch.randint(0,2,(1,256,256)),names=("C","H","W"))
frame.append_mask(mask)

newframe=T.SpatialShift((10,10))(frame)
```
* ***New feature 4*** : Added a hidden_dim argument to build_decoder
I wanted to be able to reduce the hidden dimension of the transformer to make a lightweight model, but this dimension was hardcoded in the decoder. I added the argument to be able to change it.

```
from alonet.deformable_detr.deformable_detr import DeformableDETR
from alonet.deformable_detr.backbone import Joiner


class DeformableDetrCustom(DeformableDETR):
    """Deformable Detr with custom transformer. For testing purposes only"""

    def __init__(self, *args, hidden_dim=256, return_intermediate_dec: bool = True, num_classes=91, **kwargs):
        backbone = self.build_backbone(
            backbone_name="resnet50", train_backbone=True, return_interm_layers=True, dilation=False
        )
        position_embed = self.build_positional_encoding(hidden_dim=256)
        backbone = Joiner(backbone, position_embed)
        transformer = self.build_transformer(
            hidden_dim=hidden_dim,
            dropout=0.1,
            nheads=8,
            dim_feedforward=1024,
            enc_layers=6,
            dec_layers=6,
            num_feature_levels=4,
            dec_n_points=4,
            enc_n_points=4,
            return_intermediate_dec=return_intermediate_dec,
        )

        super().__init__(backbone, transformer, *args, num_classes=num_classes, with_box_refine=False, **kwargs)

if __name__ == "__main__":
    model=DeformableDetrCustom(weights=None)
    print(model.transformer.decoder.layers[0].cross_attn.sampling_offsets) #in_features should be 256

    model2=DeformableDetrCustom(hidden_dim=128,weights=None)
    print(model2.transformer.decoder.layers[0].cross_attn.sampling_offsets) #in_features should be 128
```

_____
This pull request includes

### DIFF
--- a/alonet/deformable_detr/deformable_detr.py
+++ b/alonet/deformable_detr/deformable_detr.py
@@ -105,7 +105,6 @@ class DeformableDETR(nn.Module):
         self.background_class = num_classes if self.activation_fn == "softmax" else None
         num_classes += 1 if self.activation_fn == "softmax" else 0  # Add bg class
 
-
         # Projection for Multi-scale features
         if num_feature_levels > 1:
             num_backbone_outs = len(backbone.strides) - 1  # Ignore backbone.layer1
@@ -618,7 +617,7 @@ class DeformableDETR(nn.Module):
             n_points=dec_n_points,
         )
 
-    def build_decoder(self, hidden_dim=256, num_feature_levels=4, dec_layers: int = 6, return_intermediate_dec: bool = True):
+    def build_decoder(self, dec_layers: int = 6, return_intermediate_dec: bool = True, hidden_dim: int = 256, num_feature_levels: int = 4):
         """Build decoder layer
 
         Parameters
--- a/alonet/deformable_detr/deformable_detr.py
+++ b/alonet/deformable_detr/deformable_detr.py
@@ -618,11 +618,15 @@ class DeformableDETR(nn.Module):
             n_points=dec_n_points,
         )
 
-    def build_decoder(self, dec_layers: int = 6, return_intermediate_dec: bool = True):
+    def build_decoder(self, hidden_dim=256, num_feature_levels=4, dec_layers: int = 6, return_intermediate_dec: bool = True):
         """Build decoder layer
 
         Parameters
         ----------
+        hidden_dim : int, optional
+            Hidden dimension size, by default 256
+        num_feature_levels : int, optional
+            Number of feature levels, by default 4
         dec_layers : int, optional
             Number of decoder layers, by default 6
         return_intermediate_dec : bool, optional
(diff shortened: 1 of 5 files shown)


alonet/detr/detr.py
```diff
@@ -87,15 +92,11 @@ def __init__(
         self.background_class = self.num_classes if background_class is None else background_class
         self.num_classes += 1
 
-        hidden_dim = transformer.d_model
-        self.hidden_dim = hidden_dim
+
```
Data-Iab, line 95: 1 line is the max space

alonet/deformable_detr/deformable_detr.py
```diff
@@ -105,9 +105,7 @@ def __init__(
         self.background_class = num_classes if self.activation_fn == "softmax" else None
         num_classes += 1 if self.activation_fn == "softmax" else 0  # Add bg class
 
-        self.class_embed = nn.Linear(self.hidden_dim, num_classes)
-        self.bbox_embed = MLP(self.hidden_dim, self.hidden_dim, 4, 3)
-        self.query_embed = nn.Embedding(num_queries, self.hidden_dim * 2)
+
```
Data-Iab, line 108: 1 line is the max space

Data-Iab (approved):

PR #322, opened 2023-02-27
Generic datamodule

Dee61298:
Introducing base classes for datamodules and train pipelines (inspired by BaseDataset class).
@thibo73800 

* ***New feature 1*** : BaseDataModule class
My motivation for this class is that I kept reusing code solutions from other projects, such as the arguments, the aug/no aug train_transform structure, etc. This created quite a bit of Ctrl+C/Ctrl+V which is undesirable. My view for this class is that in the future, when creating a DataModule for a project, we inherit from the BaseDataModule class and implement only the transforms and the setup. It acts as a wrapper to the Pytorch Lightning Datamodule class, to provide all aloception users with a common code base.

* ***New feature 2*** : BaseLightningModule

Same motivation, but for training pipelines. This time, the often-reused bits are the arguments again, the optimizers, the run functions, etc.  When inheriting, the user needs to implement the model and the criterion. The user is of course free to write its own functions in the child class for more complex cases
_____
This pull request includes


thibo73800:
I think @anhtu293 & @Data-Iab  opinion are on this PR could be great to move on

Data-Iab (changes requested):
Your saving configuration is not happy with long lines, this affects the code readability in some cases, could you try to fix this ?

anhtu293:
Very good idea !
@Dee61298 in base lightning module, you shouldnt redefine the flags which are already defined by lightning (https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#trainer-flags)

Dee61298:
Closing this because I made PR #341 without the formatting issues.

PR #342, opened 2023-03-17
added .npz support to load_depth function

Dee61298:
Now you can also use this function to load depth saved as .npz. Based on the fix I made for MOTSynth dataset.

_____
This pull request includes

### DIFF
--- a/aloscene/io/depth.py
+++ b/aloscene/io/depth.py
@@ -8,12 +8,14 @@ def load_depth(path):
     Parameters
     ----------
     path: str
-        path to the disparity file. Supported format: {".npy"}. If your file is stored differently, as an
+        path to the disparity file. Supported format: {".npy",".npz"}. If your file is stored differently, as an
         alternative, you can open the file yourself and then create the Depth augmented Tensor from the depth
         data.
     """
     if path.endswith(".npy"):
         return np.load(path)
+    elif path.endswith(".npz"):
+        return np.load(path)["arr_0"]
     else:
         raise ValueError(
             f"Unknown extension for depth file: {path}. As an alternative you can load the file manually\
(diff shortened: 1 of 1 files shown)


aloscene/io/depth.py
```diff
@@ -8,12 +8,14 @@ def load_depth(path):
...
     path: str
-        path to the disparity file. Supported format: {".npy"}. If your file is stored differently, as an
+        path to the disparity file. Supported format: {".npy",".npz"}. If your file is stored differently, as an
         alternative, you can open the file yourself and then create the Depth augmented Tensor from the depth
         data.
     """
     if path.endswith(".npy"):
         return np.load(path)
+    elif path.endswith(".npz"):
+        return np.load(path)["arr_0"]
```
Data-Iab, line 18: This key-specific, not all .npz have `"arr_0"` as a key

Data-Iab (changes requested):

thibo73800 (approved):

thibo73800 (approved):

aloscene/io/depth.py
```diff
@@ -8,12 +8,14 @@ def load_depth(path):
...
     path: str
-        path to the disparity file. Supported format: {".npy"}. If your file is stored differently, as an
+        path to the disparity file. Supported format: {".npy",".npz"}. If your file is stored differently, as an
         alternative, you can open the file yourself and then create the Depth augmented Tensor from the depth
         data.
     """
     if path.endswith(".npy"):
         return np.load(path)
+    elif path.endswith(".npz"):
+        return list(np.load(path).values())
```
thibo73800, line 18: Not sure we want to return the list of delement here. I think we want to retrun only one depth element assuming only one depth is inside the npz. Somethine link
```
content = np.load(path)
return content[content.keys()[0]]
```

Data-Iab (approved):

PR #355, opened 2023-04-28
Update Dockerfile

Aurelien-VB:
Fix typing error in dockerfile


Aurelien-VB:
It is actually used line 52. I'm not even sure how this docker file work with the typing error.

PR #356, opened 2023-05-16
use p=1 by default for compose

Aurelien-VB:
Right now Compose has a probability of 0.5 by default causing very unintuive behavior (skipping all transforms half of the time).

I propose to replace to 1 by default with the option to change it.


Aurelien-VB:
I also removed the _p_ argument in RandomSelect. The problem was that if we pass  _p_ it will be consumed by this and wouldn't go to the inherited class, causing it to overwrite _self.p_ to 0.5

